PR #2, opened 2019-07-16
Server Factory

amandaratio:
1. Added methods to handle Mqtt `Ping` and `Disconnect` Packets after `Connect` and `Connack` exchange

2. Added tests for `PingReq` & `PingResp` exchange as well as `Disconnect`

### DIFF
--- a/src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
+++ b/src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
@@ -518,9 +518,15 @@ public final class MqttServerFactory implements StreamFactory
 
         private void doMqttDisconnect()
         {
-            final MqttDisconnectFW disconnect = mqttDisconnectRW.wrap(writeBuffer,  0, writeBuffer.capacity())
+            OctetsFW properties = new OctetsFW.Builder()
+                .wrap(writeBuffer, 0, writeBuffer.capacity())
+                .build();
+
+            final MqttDisconnectFW disconnect = mqttDisconnectRW.wrap(writeBuffer,  DataFW.FIELD_OFFSET_PAYLOAD, writeBuffer.capacity())
                 .packetType(0xD0)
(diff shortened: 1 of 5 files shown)


src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -18,6 +18,7 @@
 
 import static java.util.Objects.requireNonNull;
 
+import org.reaktivity.nukleus.mqtt.internal.types.codec.*;
 import org.reaktivity.nukleus.mqtt.internal.types.stream.*;
```
jfallows: Avoid wildcard imports, use explicit imports instead.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -345,6 +358,31 @@ private void onMqttConnect(
...
+        {
+        }
+
+        private void onMqttPublish()
+        {
+        }
+
+        private void onMqttDisconnect(
+            MqttDisconnectFW disconnect
+        )
```
jfallows: Formatting, trailing `)` on next line.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -468,11 +553,27 @@ private int decodePacketType(
...
+                    break;
+                case 0xA2:
+                    /* onUnsubscribe */
+                    break;
+                case 0xE0:
+                    /* decode reason code */
+                    break;
+                default:
+                    doAbort(decodeTraceId);
+                    break;
```
jfallows, line 600: We typically use lowercase for hex literals, so `0xc0` rather than `0xC0`.

jfallows (changes requested):

jfallows (approved):

PR #3, opened 2019-07-20
Server Factory

amandaratio:


src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -18,17 +18,7 @@
...
-import org.reaktivity.nukleus.mqtt.internal.types.codec.MqttConnackFW;
-import org.reaktivity.nukleus.mqtt.internal.types.codec.MqttPingReqFW;
-import org.reaktivity.nukleus.mqtt.internal.types.codec.MqttPingRespFW;
-import org.reaktivity.nukleus.mqtt.internal.types.codec.MqttSubscribeFW;
-import org.reaktivity.nukleus.mqtt.internal.types.codec.MqttSubackFW;
-import org.reaktivity.nukleus.mqtt.internal.types.codec.MqttUnsubscribeFW;
-import org.reaktivity.nukleus.mqtt.internal.types.codec.MqttUnsubackFW;
-import org.reaktivity.nukleus.mqtt.internal.types.codec.MqttPublishFW;
-import org.reaktivity.nukleus.mqtt.internal.types.codec.MqttDisconnectFW;
+import org.reaktivity.nukleus.mqtt.internal.types.codec.*;
```
jfallows: Avoid wildcard imports, replace with explicit imports, eliminating diff.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -472,136 +518,192 @@ private void doSignal(
...
-            final int packetType = mqttPacket.packetType();
+            final MqttPacketFW mqttPacket = mqttPacketRO.tryWrap(buffer, offset, offset + length);
+            final int packetType = mqttPacket == null ? 0 : mqttPacket.packetType();
 
             switch (packetType)
             {
                 case 0xc0:
                     final MqttPingReqFW ping = mqttPingReqRO.tryWrap(buffer, offset, offset + length);
                     onMqttPingReq(ping);
+                    this.decodeState = this::decodeSession;
```
jfallows: Already in the right `decodeState` no need to reassign same `this::decodeSession` value.
Check other cases too.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -472,136 +518,192 @@ private void doSignal(
...
                 case 0xa2:
-                    /* onMqttUnsubscribe */
+                    final MqttUnsubscribeFW unsubscribe = mqttUnsubscribeRO.tryWrap(buffer, offset, offset + length);
+                    onMqttUnsubscribe(unsubscribe);
+                    this.decodeState = this::decodeSession;
                     break;
                 case 0xe0:
-                    /* onMqttDisconnect decode reason code */
+                    final MqttDisconnectFW disconnect = mqttDisconnectRO.tryWrap(buffer, offset, offset + length);
+                    this.decodeState = this::decodeEnd;
```
jfallows: Move `decodeState` assignment to `onMqttDisconnect` instead.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -366,31 +368,58 @@ private void onSignal(
...
+        private void onMqttPublish(
+            MqttPublishFW publish)
         {
         }
 
         private void onMqttDisconnect(
             MqttDisconnectFW disconnect)
         {
-
+            /* process reason code */
```
jfallows, line 510: Call `doMqttDisconnect(...)` and `doEnd(...)` here.
Eliminates need for `decodeEnd` ?

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -472,136 +518,192 @@ private void doSignal(
...
+            if (!protocol || packet.protocolVersion() != 5)
+            {
+                return 132;
+            }
+            else if (((packet.flags() >> 1) & 1) == 0)
+            {
+                return 129;
+            }
+
+            return 0;
```
jfallows: Inline this method to `onMqttConnect` instead (only used once).

Inline `if` condition, and reverse comparison to simplify `null` check.
Also, simplify logic to single-return.

```
int reasonCode = 0x00;

final String protocolName = packet.protocolName().asString();
if (!"MQTT".equals(protocolName) || packet.protocolVersion() != 5)
{
    reasonCode = 0x84; // unsupported protocol version
}
else if ((packet.flags() & 0b0000_0001) != 0b0000_0000)
{
    reasonCode = 0x81; // malformed packet
}

return reasonCode;
```
I think the intention of the flags check is `3.1.2.3 Connect Flags` which states that `the Server MUST validate that the reserved flag in the CONNECT packet is set to 0`.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -472,136 +518,192 @@ private void doSignal(
...
-        private boolean isValidConnectPacket(
+        private int getConnackReasonCode(
             MqttConnectFW packet)
         {
-            return packet.protocolName().toString().equals("MQTT")
-                && packet.protocolVersion() == 0x05;
+            if (packet == null)
+            {
+                return 128;
+            }
```
jfallows: We shouldn't have to defend against null here, since decode loop will buffer until packet is complete before calling `onMqttConnect(...)`.

src/main/reaktivity/protocol.idl
```diff
@@ -160,6 +171,46 @@
             octets[propertiesLength] properties;
         }
 
+        struct Puback
```
jfallows: `Puback` => `MqttPuback`

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -323,6 +314,17 @@ else if (payload != null)
             }
```
jfallows, line 371: This is where the `decode` loop should go.

In fact, given that all `MQTT` control packets have the same general layout, it would make sense to add the following structure to `protocol.idl`.
```
struct MqttPacketFixedHeader
{
    uint8 typeAndFlags;
    uint8 remainingLength;
}
```
and use it here to detect when there are sufficient bytes to decode a full `MqttPacket` including `remainingLength`.

Note: all `Mqtt*` packet structures in `idl` can `extend` `MqttPacketFixedHeader` to save repeating these common fields in each case.

If there are not sufficient bytes delivered to complete a successful `tryWrap` of `MqttPacketFixedHeader` or if there are not sufficient `remainingBytes`, then the partial bytes received so far should be copied to a buffer slot so they can be retained until the next `DATA` frame arrives.

When the next `DATA` frame arrives, if there is a `bufferSlot` in use, then the new `payload` bytes should be copied after the existing bytes in the `bufferSlot`, so you'll also need to track `slotLimit` and update it each time you copy bytes into the `bufferSlot` to know how much is in use.

Then, the `decodeState.decode(buffer, offset, length)` call will use the `bufferSlot` buffer instead of the `DataFW` `payload` buffer.

Note: there may be more than one packet to decode so you need to advance through the buffer until the `decoderState` is no longer making progress.

When the buffer has been decoded as far as possible, then the remaining bytes need to be copied to the beginning of the buffer slot.

If all the bytes have been consumed, then the `bufferSlot` (if any) should be returned to the `bufferPool`.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -366,31 +368,58 @@ private void onSignal(
...
+            MqttSubscribeFW subscribe)
         {
+            int topics = subscribe == null ? 0 : subscribe.topics().sizeof();
+
+            DirectBuffer buffer = subscribe.topics().buffer();
+            int offset = subscribe.topics().offset();
+            int length = subscribe.topics().sizeof();
+
+            MqttSubscriptionTopicsFW subs = new MqttSubscriptionTopicsFW()
+                .tryWrap(buffer, offset, offset + length);
```
jfallows: Use preallocated `MqttSubsbcriptionTopicsFW` instead of calling `new` here.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -405,6 +434,23 @@ private void doBegin(
...
+            final DataFW data = dataRW.wrap(writeBuffer, 0, writeBuffer.capacity())
+                .routeId(routeId)
+                .streamId(replyId)
+                .trace(supplyTraceId.getAsLong())
+                .groupId(0)
+                .padding(replyPadding)
+                .payload(buffer, offset, sizeOf)
+                .build();
+
+            receiver.accept(data.typeId(), data.buffer(), data.offset(), data.sizeof());
```
jfallows: Suggest renaming `receiver` to `network` as a parallel to `onNetwork`.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -472,136 +518,192 @@ private void doSignal(
             receiver.accept(signal.typeId(), signal.buffer(), signal.offset(), signal.sizeof());
         }
 
-        private void doMqttConnack()
+        private void doMqttConnack(
+            int reasonCode)
         {
             OctetsFW properties = new OctetsFW.Builder()
                 .wrap(writeBuffer, 0, writeBuffer.capacity())
                 .build();
```
jfallows, line 614: Use preallocated `OctetsFW.Builder()` instead of calling `new` here.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -472,136 +518,192 @@ private void doSignal(
...
         }
 
-        private void doMqttSuback()
+        private void doMqttSuback(
+            int subscriptions)
         {
+            OctetsFW reasonCodes = new OctetsFW.Builder()
+                .wrap(writeBuffer, 0, writeBuffer.capacity())
+                .put(new byte[] {0x00})
+                .build();
```
jfallows, line 644: Use preallocated `OctetsFW.Builder()` instead of calling `new` here.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -472,136 +518,192 @@ private void doSignal(
...
         }
 
-        private void doMqttUnsuback()
+        private void doMqttUnsuback(
+            int subsriptions)
         {
+            OctetsFW reasonCodes = new OctetsFW.Builder()
+                .wrap(writeBuffer, 0, writeBuffer.capacity())
+                .put(new byte[] {0x00})
+                .build();
```
jfallows, line 662: Use preallocated `OctetsFW.Builder()` instead of calling `new` here.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -472,136 +518,192 @@ private void doSignal(
...
+            doData(unsuback.buffer(), unsuback.offset(), unsuback.sizeof());
         }
 
-        private void doMqttDisconnect()
+        private void doMqttDisconnect(
+            int reasonCode)
         {
             OctetsFW properties = new OctetsFW.Builder()
                 .wrap(writeBuffer, 0, writeBuffer.capacity())
                 .build();
```
jfallows, line 679: Use preallocated `OctetsFW.Builder()` instead of calling `new` here.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -472,136 +518,192 @@ private void doSignal(
...
+            doData(disconnect.buffer(), disconnect.offset(), disconnect.sizeof());
+        }
 
-            receiver.accept(data.typeId(), data.buffer(), data.offset(), data.sizeof());
+        private boolean setReservedSubscriptionOptionsBits(
+            int options)
+        {
+            return ((options >> 8) & 1) != 0
+                && ((options >> 7) & 1) != 0;
         }
```
jfallows, line 698: Recommend you change this logic to verify correctness, rather than checking for failure.
Also, if this will only be validated in `onMqttSubscribe` then no need to separate it out into another method.

src/main/reaktivity/protocol.idl
```diff
@@ -111,7 +122,7 @@
         struct MqttPingResp
         {
             uint8 packetType = 0xd0;
```
jfallows: Consider moving packet type fixed value to `enum` instead, requires https://github.com/reaktivity/nukleus-maven-plugin/pull/89.

```
enum MqttPacketType (uint8)
{
    CONNECT (1),
    CONNACK (2),
    PUBLISH (3),
    PUBACK (4),
    PUBREC (5),
    PUBREL (6),
    PUBCOMP (7),
    SUBSCRIBE (8),
    SUBACK (9),
    UNSUBSCRIBE (10),
    UNSUBACK (11),
    PINGREQ (12),
    PINGRESP (13),
    DISCONNECT (14),
    AUTH (15)
}
```

jfallows (changes requested):

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -366,31 +414,98 @@ private void onSignal(
...
+            while (length > 0)
+            {
+                MqttTopicFW topic = mqttTopicRO
+                    .tryWrap(buffer, offset, offset + length);
+
+                offset += topic.sizeof();
+                length -= topic.sizeof();
+
+                topicsLength++;
+            }
```
jfallows, line 497: Let's change this to be limit-based instead of length-based.

Suggest the following:
```
            final OctetsFW topicFilters = unsubscribe.topicFilters();
            final DirectBuffer buffer = topicFilters.buffer();
            final int limit = topicFilters.limit();
            final int offset = topicFilters.offset();

             int topics = 0;
             MqttTopicFW topic;
             for (int progress = offset; progress < limit; progress = topic.limit())
             {
                topic = mqttTopicRO.tryWrap(buffer, progress, limit);
                if (topic == null)
                {
                    break;
                }
                topics++;
             }
```

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -469,135 +601,160 @@ private void doSignal(
...
+                final MqttPublishFW publish = mqttPublishRO.wrap(buffer, offset, limit);
+                onMqttPublish(publish);
+                consumed = publish.limit();
+                break;
+            default:
+                doReset(decodeTraceId);
+                break;
             }
 
             return consumed;
```
jfallows: Rename `consumed` to `progress`.

src/main/reaktivity/protocol.idl
```diff
@@ -38,24 +46,28 @@
...
+        struct MqttSubscriptionTopic
         {
-            string topic;
+            string16 topicFilter;
             uint8 options;
         }
 
+        struct MqttTopic
+        {
+            string16 topicFilter;
```
jfallows: Rename `topicFilter` to `filter` since it is in the context of `MqttTopic` already.

src/test/java/org/reaktivity/nukleus/mqtt/internal/ConnectionIT.java
```diff
@@ -52,7 +52,7 @@
     @Test
     @Specification({
         "${route}/server/controller",
-        "${client}/connect/client"})
+        "${client}/connect/successful/complete/client"})
```
jfallows: Move `successful/complete` to `successful`.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -313,13 +322,52 @@ private void onData(
...
+
+                    final MqttPacketType packetType = MqttPacketType.valueOf(packet.typeAndFlags() >> 4);
+                    offset += decodeState.decode(packetType, buffer, offset, limit);
+                }
+
+                if (offset < limit)
+                {
+                    if (bufferSlot == NO_SLOT)
+                    {
+                        bufferSlot = bufferPool.acquire(streamId);
```
jfallows: Rename `bufferSlot` to `slotIndex` so that all slot-related variables start with `slot` prefix for readability.

jfallows (changes requested):

jfallows (dismissed):

jfallows (approved):

PR #6, opened 2019-07-26
High Level Streams

amandaratio:
1. initiates stream for `PUBLISH`
2. initiates stream for `SUBSCRIBE`
3. Handles stream for `SUBSCRIBE` with proceeding `PUBLISH`

### DIFF
--- a/src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
+++ b/src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
@@ -640,7 +640,7 @@ public final class MqttServerFactory implements StreamFactory
                             decodeTraceId, dataEx.buffer(), dataEx.offset(), dataEx.sizeof(), payload);
 
                         correlations.put(newReplyId, serverStream);
-                        mqttSubscriptionStreams.put(topicName, serverStream);
+                        mqttPublishStreams.put(topicName, serverStream);
                     }
                 }
             }
--- a/src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
+++ b/src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
@@ -93,6 +93,8 @@ public final class MqttServerFactory implements StreamFactory
     private final ResetFW.Builder resetRW = new ResetFW.Builder();
     private final SignalFW.Builder signalRW = new SignalFW.Builder();
 
+    private final MqttDataExFW mqttDataExRO = new MqttDataExFW();
+
     private final MqttBeginExFW.Builder mqttBeginExRW = new MqttBeginExFW.Builder();
     private final MqttDataExFW.Builder mqttDataExRW = new MqttDataExFW.Builder();
     private final MqttEndExFW.Builder mqttEndExRW = new MqttEndExFW.Builder();
@@ -125,6 +127,7 @@ public final class MqttServerFactory implements StreamFactory
     private final MqttSubackFW.Builder mqttSubackRW = new MqttSubackFW.Builder();
     private final MqttUnsubscribeFW.Builder mqttUnsubscribeRW = new MqttUnsubscribeFW.Builder();
     private final MqttUnsubackFW.Builder mqttUnsubackRW = new MqttUnsubackFW.Builder();
+    private final MqttPublishFW.Builder mqttPublishRW = new MqttPublishFW.Builder();
 
     private final RouteManager router;
     private final MutableDirectBuffer writeBuffer;
@@ -530,13 +533,14 @@ public final class MqttServerFactory implements StreamFactory
 
                     if (route != null)
                     {
-                        final MqttServerStream serverStream = new MqttServerStream();
-
                         final long newRouteId = route.correlationId();
                         final long newInitialId = supplyInitialId.applyAsLong(newRouteId);
                         final long newReplyId = supplyReplyId.applyAsLong(newInitialId);
 
                         final MessageConsumer newTarget = router.supplyReceiver(newInitialId);
+
+                        final MqttServerStream serverStream = new MqttServerStream(this, newTarget,
+                            newRouteId, newInitialId, newReplyId);
 
                         final MqttBeginExFW beginEx = mqttBeginExRW
                             .wrap(extBuffer, 0, extBuffer.capacity())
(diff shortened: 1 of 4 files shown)


src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -275,7 +281,8 @@ private RouteFW resolveTarget(
         private DecoderState decodeState;
         private int slotIndex = NO_SLOT;
         private int slotLimit;
-        private HashMap<String, MqttServerStream> mqttStreams;
+        private HashMap<String, MqttServerStream> mqttSubscriptionStreams;
+        private HashMap<String, MqttServerStream> mqttPublishStreams;
```
jfallows: Please change these types to `private final Map<String, MqttServerStream>` instead.
Suggest renaming the fields to `subscribers` and `publishers`.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -586,6 +594,61 @@ private void onMqttUnsubscribe(
         private void onMqttPublish(
             MqttPublishFW publish)
         {
+            if (publish != null)
+            {
```
jfallows: Do we need to test for `null` here or is `publish` guaranteed to always be `non-null`?

jfallows (changes requested):

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -275,7 +284,8 @@ private RouteFW resolveTarget(
         private DecoderState decodeState;
         private int slotIndex = NO_SLOT;
         private int slotLimit;
-        private HashMap<String, MqttServerStream> mqttStreams;
+        private Map<String, MqttServerStream> subscribers;
+        private Map<String, MqttServerStream> publishers;
```
jfallows, line 288: Please mark these as `final` and move up with other `final` fields.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -872,19 +955,87 @@ private void onApplication(
...
+            final WindowFW window = windowRW.wrap(writeBuffer, 0, writeBuffer.capacity())
+                .routeId(routeId)
+                .streamId(replyId)
+                .trace(supplyTraceId.getAsLong())
+                .credit(bufferPool.slotCapacity())
+                .padding(0)
+                .groupId(0)
+                .build();
 
+            application.accept(window.typeId(), window.buffer(), window.offset(), window.sizeof());
```
jfallows, line 967: Recommend hoisting the above logic into `doWindow` method on `MqttServer`.

When we receive the `BEGIN` reply here, it should trigger the sending of `SUBACK` frame if the initial `MqttServerStream` was created for `SUBSCRIBE`, right?

So that would be `server.onSubscribed(...)` and initially we can assume just one topic filter per `SUBSCRIBE`.

Next step would be to defer sending the `SUBACK` frame until all `reason` codes are collected from multiple application streams, triggered by sending parallel application `BEGIN` streams in reaction to the same inbound `SUBSCRIBE` frame.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -872,19 +955,87 @@ private void onApplication(
...
+        }
+
+        private void onData(
+            DataFW data)
+        {
+            OctetsFW ex = data.extension();
```
jfallows, line 973: `final` and rename to `extension`.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -872,19 +955,87 @@ private void onApplication(
...
+
+            final DirectBuffer exBuffer = data.extension().buffer();
+            final int exOffset = data.extension().offset();
+            final int exLength = data.extension().sizeof();
```
jfallows, line 977: Use `extension` variable instead of calling `data.extension()` multiple times.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -842,8 +901,24 @@ private int decodeSession(
...
+        private long replyId;
+
+        MqttServerStream(
+            MqttServer network,
+            MessageConsumer application,
+            long routeId,
+            long initialId,
+            long replyId)
         {
+            this.receiver = network;
```
jfallows, line 917: Rename to `this.server = server`.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -872,19 +955,87 @@ private void onApplication(
...
+            final DataFW dataFW = dataRW.wrap(writeBuffer, 0, writeBuffer.capacity())
+                .routeId(receiver.routeId)
+                .streamId(receiver.replyId)
+                .trace(supplyTraceId.getAsLong())
+                .groupId(0)
+                .padding(0)
+                .payload(publishBuffer, publishOffset, publishLength)
+                .build();
+
+            this.receiver.network.accept(dataFW.typeId(), dataFW.buffer(), dataFW.offset(), dataFW.sizeof());
```
jfallows, line 1008: This becomes `server.network.accept(...)`

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -872,19 +955,87 @@ private void onApplication(
...
         private void onReset(
             ResetFW reset)
         {
+        }
+
+        private void doBegin(
+            MessageConsumer target,
+            long routeId,
+            long streamId,
+            long traceId)
```
jfallows, line 1030: Most of these arguments are already available as member variables and can be removed.
Please apply same feedback to other `doXXX` methods on `MqttServerStream`.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
jfallows: Also refactor above code from `983` onwards into `server.doMqttPublish(...)`

jfallows (changes requested):

PR #7, opened 2019-09-18
High Level Streams

dknoma:
1. initiates stream for PUBLISH
2. initiates stream for SUBSCRIBE
3. Handles stream for SUBSCRIBE with proceeding PUBLISH

### DIFF
--- a/src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
+++ b/src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
@@ -581,7 +581,7 @@ public final class MqttServerFactory implements StreamFactory
 
             for (int i = 0; i < reasonCodeCount; i++)
             {
-                if ((nullRoutesMask & 1 << i) == 1)
+                if ((nullRoutesMask & (1 << i)) > 0)
                 {
                     subscription.setReasonCode(i, 0x8f, this::doMqttSuback);
                 }
--- a/src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
+++ b/src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
@@ -21,11 +21,9 @@ import static java.util.Objects.requireNonNull;
 import static org.reaktivity.nukleus.buffer.BufferPool.NO_SLOT;
 
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.function.Consumer;
 import java.util.function.LongSupplier;
 import java.util.function.LongUnaryOperator;
 import java.util.function.ToIntFunction;
@@ -495,6 +493,7 @@ public final class MqttServerFactory implements StreamFactory
             final DirectBuffer buffer = topicFilters.buffer();
             final int limit = topicFilters.limit();
             final int offset = topicFilters.offset();
+            final int packetId = subscribe.packetId();
             int reasonCodeCount = 0;
             int subscriptionId = 0;
             int nullRoutesMask = 0;
@@ -536,9 +535,9 @@ public final class MqttServerFactory implements StreamFactory
                     //      - need to be able to map the correct reason codes to the correct streams
                     //      - defer reason code assignment to WINDOW
                     // need to map reason code -> stream in order of when they were created, not when they come in
-                    // 	- ie. streams: 0, 1, 2, 3
+                    //  - ie. streams: 0, 1, 2, 3
                     //
-                    // 		even if 1 sends WINDOW before 0, should sitll map to list[1], not list[0]
+                    //      even if 1 sends WINDOW before 0, should sitll map to list[1], not list[0]
                     //      would streams be assigned a number to indicate their reasoncodesIndex?
                     // int reasonCode = 0x8f;  // 0x8F - Topic Filter invalid
                     final String topicFilter = subscription.topicFilter().asString();
@@ -553,7 +552,7 @@ public final class MqttServerFactory implements StreamFactory
                         final MessageConsumer newTarget = router.supplyReceiver(newInitialId);
 
                         final MqttServerStream serverStream = new MqttServerStream(this, newTarget,
-                            newRouteId, newInitialId, newReplyId, ackIndex, subscriptionId);
+                            newRouteId, newInitialId, newReplyId, ackIndex, packetId);
 
                         serverStream.doMqttBeginEx(decodeTraceId, topicFilter, subscriptionId);
 
@@ -574,8 +573,8 @@ public final class MqttServerFactory implements StreamFactory
             // TODO - For each non-routable server stream, explicitly set the failure reason code at the corresponding ackIndex.
             //        If all server streams were non-routable, then onMqttSubscribe should respond immediately because all reason
             //        codes are known.
-            Subscription subscription = new Subscription(reasonCodeCount);
-            subscriptionsByPacketId.put(subscriptionId, subscription);
+            Subscription subscription = new Subscription(this, reasonCodeCount);
+            subscriptionsByPacketId.put(packetId, subscription);
 
             System.out.printf("reasonCodeCount: %d\n", reasonCodeCount);
 
@@ -583,10 +582,10 @@ public final class MqttServerFactory implements StreamFactory
             {
                 if ((nullRoutesMask & (1 << i)) > 0)
                 {
-                    subscription.setReasonCode(i, 0x8f, this::doMqttSuback);
+                    subscription.onSubscribeFailed(packetId, i);
                 }
             }
-            correlations.forEach((id, stream) -> stream.subscription = subscription);
+            subscribers.forEach((filter, stream) -> stream.subscription = subscription);
 
         }
 
@@ -681,7 +680,6 @@ public final class MqttServerFactory implements StreamFactory
             router.setThrottle(replyId, this::onNetwork);
         }
 
-        // TODO: can change to Flyweight payload?
         private void doData(
             Flyweight payload)
         {
@@ -826,7 +824,8 @@ public final class MqttServerFactory implements StreamFactory
         }
 
         private void doMqttSuback(
-            byte[] subscriptions)
+            byte[] subscriptions,
+            int packetId)
         {
             OctetsFW reasonCodes = octetsRW
                 .wrap(writeBuffer, 0, writeBuffer.capacity())
@@ -837,6 +836,7 @@ public final class MqttServerFactory implements StreamFactory
                 .typeAndFlags(0x90)
                 .remainingLength(reasonCodes.sizeof() + 1)
                 .propertiesLength(0x00)
+                .packetId(packetId)
                 .reasonCodes(reasonCodes)
                 .build();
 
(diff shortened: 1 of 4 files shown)


dknoma:
> Please mark these as `final` and move up with other `final` fields.

> Use extension variable instead of calling data.extension() multiple times.

> Also refactor above code from 983 onwards into server.doMqttPublish(...)

> Most of these arguments are already available as member variables and can be removed.
Please apply same feedback to other doXXX methods on MqttServerStream.

Made these requested changes from the other PR

dknoma:
>  Recommend hoisting the above logic into doWindow method on MqttServer.
> 
> When we receive the BEGIN reply here, it should trigger the sending of SUBACK frame if the initial MqttServerStream was created for SUBSCRIBE, right?
> 
> So that would be server.onSubscribed(...) and initially we can assume just one topic filter per SUBSCRIBE.
> 
> Next step would be to defer sending the SUBACK frame until all reason codes are collected from multiple application streams, triggered by sending parallel application BEGIN streams in reaction to the same inbound SUBSCRIBE frame.

So just remove this code from `onBegin`, but `doWindow` already is sending a `WINDOW` to the `network` so is there any extra work that needs to be done? or just creating a new `onSubscribed` for the `onBegin`?

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -505,55 +530,90 @@ private void onMqttSubscribe(
...
                             .typeId(mqttTypeId)
                             .role(r -> r.set(MqttRole.RECEIVER))
                             .clientId("client")
                             .topic(topicFilter)
                             .subscriptionId(subscriptionId)
                             .build();
 
-                        serverStream.doMqttBeginEx(newTarget, newRouteId, newInitialId,
-                            decodeTraceId, beginEx.buffer(), beginEx.offset(), beginEx.sizeof());
+                        serverStream.doMqttBeginEx(decodeTraceId, beginEx);
```
jfallows: Suggest moving the `MqttBeginEx` construction into the `serverStream.doMqttBegin(...)` instead.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -505,58 +523,75 @@ private void onMqttSubscribe(
                     }
                 }
 
-                for (int progress = offset; progress < limit; progress = subscription.limit())
+                for (int progress = offset, index = 0; progress < limit; progress = subscription.limit(), index++)
```
jfallows: Suggest creating an `MqttSubscription` outside the group and keeping it in a map keyed by `packetId`, then passing `packetId` and `ackIndex` to each `MqttServerStream` that is created.

For each non-routable server stream, explicitly set the failure reason code at the corresponding `ackIndex`.

If all server streams were non-routable, then `onMqttSubscribe` should respond immediately because all reason codes are known.

Otherwise, when each `MqttServerStream` gets an initial throttle frame:
  `RESET` should trigger `onSubscribeFailed(packetId, ackIndex)`, and
  `WINDOW` should trigger `onSubscribeSucceeded(packetId, ackIndex)`

`onSubscribeFailed` and `onSubscribeSucceeded` would behave similarly, choosing to report a failure or success `reasonCode` for the `ackIndex`, and detecting when all reason codes had reported in, triggering delivery of the `SUBACK` with complete list of reason codes back to the client.

jfallows (changes requested):

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -221,6 +232,7 @@ private MessageConsumer newReplyStream(
         MessageConsumer newStream = null;
         if (reply != null)
         {
+            router.setThrottle(replyId, reply::onApplication);
```
jfallows: No need for throttle on server reply, that one is being read.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -261,6 +273,9 @@ private RouteFW resolveTarget(
         private final long routeId;
         private final long initialId;
         private final long replyId;
+        private final Map<String, MqttServerStream> subscribers;
+        private final Map<String, MqttServerStream> publishers;
+        private final Map<Integer, Subscription> subscriptionsByPacketId;
```
jfallows: Use `Int2ObjectHashMap<Subscription>` instead of `Map<Integer, Subscription>` to avoid auto-boxing.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -505,56 +524,54 @@ private void onMqttSubscribe(
...
 
-                        final MqttBeginExFW beginEx = mqttBeginExRW
-                            .wrap(writeBuffer, BeginFW.FIELD_OFFSET_EXTENSION, writeBuffer.capacity())
-                            .typeId(mqttTypeId)
-                            .role(r -> r.set(MqttRole.RECEIVER))
-                            .clientId("client")
-                            .topic(topicFilter)
-                            .subscriptionId(subscriptionId)
-                            .build();
+                        final MqttServerStream serverStream = new MqttServerStream(this, newTarget,
```
jfallows: Consider moving `MqttServerStream` to an inner class of `MqttServer`, no need to explicitly pass `this`.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -505,56 +524,54 @@ private void onMqttSubscribe(
...
+                            newRouteId, newInitialId, newReplyId, ackIndex, packetId);
 
-                        serverStream.doMqttBeginEx(newTarget, newRouteId, newInitialId,
-                            decodeTraceId, beginEx.buffer(), beginEx.offset(), beginEx.sizeof());
+                        router.setThrottle(newInitialId, serverStream::onApplication);
```
jfallows: This should probably be done inside `serverStream.doMqttBeginEx` instead.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -477,7 +494,10 @@ private void onMqttSubscribe(
             final DirectBuffer buffer = topicFilters.buffer();
             final int limit = topicFilters.limit();
             final int offset = topicFilters.offset();
-            ArrayList<Byte> reasonCodes = new ArrayList<>();
+            final int packetId = subscribe.packetId();
+            int reasonCodeCount = 0;
+            int subscriptionId = 0;
+            int nullRoutesMask = 0;
```
jfallows: Rename `nullRoutesMask` to `unrouteableMask`.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -837,10 +931,93 @@ private int decodeSession(
...
+            MqttServer server,
+            int reasonCodesCount)
+        {
+            this.server = server;
+            this.reasonCodes = new ArrayList<>();
+            for (int i = 0; i < reasonCodesCount; i++)
+            {
+                reasonCodes.add((byte) 0x8f);
+            }
+            this.fullReasonCodesMask = 1 << (reasonCodesCount - 1);
```
jfallows: Let's change this to 2 masks; one `ackMask` to say which subscribes have acked, and one `successMask` to say which subscribes were successful.

The initial value of the `ackMask` should match `unrouteableMask` from `onMqttSubscribe` since those ones are complete, and unsuccessful. The initial value of `successMask` is `0`.

The last pending subscribe has been acked when the bit count of `ackMask` reaches `reasonCodesCount`, which we can probably rename to `ackCount`.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -837,10 +931,93 @@ private int decodeSession(
...
+            int ackIndex)
+        {
+            final int bit = 1 << ackIndex;
+            reasonCodes.set(ackIndex, (byte) 0x00);
+            reasonCodesMask |= bit;
+            if ((reasonCodesMask & fullReasonCodesMask) == fullReasonCodesMask)
+            {
+                doMqttSuback(packetId);
+            }
+        }
```
jfallows: Both `onSubscribeFailed` and `onSubscribeSucceeded` should defer to `onSubscribeCompleted(..., ackIndex, reasonCode)`, passing `reasonCode` `0x00` for succeeded and `0x8f` for failed.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -837,10 +931,93 @@ private int decodeSession(
...
+        {
+            byte[] subackReasonCodes = new byte[reasonCodes.size()];
+            for (int i = 0; i < reasonCodes.size(); i++)
+            {
+                subackReasonCodes[i] = reasonCodes.get(i);
+            }
+
+            server.doMqttSuback(subackReasonCodes, packetId);
+        }
+    }
```
jfallows: This can inline to `onSubscribeCompleted` inside the test for `if (Integer.bitCount(acksMask) == ackCount)`.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -837,10 +931,93 @@ private int decodeSession(
...
+        private void doMqttSuback(
+            int packetId)
+        {
+            byte[] subackReasonCodes = new byte[reasonCodes.size()];
+            for (int i = 0; i < reasonCodes.size(); i++)
+            {
+                subackReasonCodes[i] = reasonCodes.get(i);
+            }
+
+            server.doMqttSuback(subackReasonCodes, packetId);
```
jfallows: If you make `Subscription` an inner class of `MqttServer`, then no need to track `server`.

jfallows (changes requested):

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -505,56 +524,52 @@ private void onMqttSubscribe(
...
+            subscriptionsByPacketId.put(packetId, subscription);
+
+            for (int i = 0; i < reasonCodeCount; i++)
             {
-                subackReasonCodes[i] = reasonCodes.get(i);
+                if ((unrouteableMask & (1 << i)) > 0)
+                {
+                    subscription.onSubscribeFailed(packetId, i, 0x8f);
+                }
             }
```
jfallows: This loop is unnecessary, right?

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -505,56 +524,52 @@ private void onMqttSubscribe(
...
-            doMqttSuback(subackReasonCodes);
+            subscribers.forEach((filter, stream) -> stream.subscription = subscription);
```
jfallows: Prefer to pre-create the `Subscription`, pass to each `MqttServerStream`, then after all streams processed, update `subscription.ackCount` to `reasonCodeCount` and `subscription.ackMask` to `unrouteableMask`.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -835,72 +927,217 @@ private int decodeSession(
...
-                onReset(reset);
-                break;
+                this.ackMask = unrouteableMask;
+                this.ackCount = ackCount;
+                this.successMask = 0;
+                this.reasonCodes = new ArrayList<>();
+                for (int i = 0; i < ackCount; i++)
+                {
+                    reasonCodes.add((byte) 0x8f);
+                }
```
jfallows: We don't need `reasonCodes` any more, since they are derivable from `successMask`, right?

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -835,72 +927,217 @@ private int decodeSession(
...
+                int packetId)
+            {
+                if (Integer.bitCount(ackMask) == ackCount)
+                {
+                    byte[] subackReasonCodes = new byte[reasonCodes.size()];
+                    for (int i = 0; i < reasonCodes.size(); i++)
+                    {
+                        subackReasonCodes[i] = reasonCodes.get(i);
+                    }
+                    doMqttSuback(subackReasonCodes, packetId);
```
jfallows: Suggest passing `ackMask` and `successMask` instead of `subackReasonCodes` array, then `doMqttSuback` will loop the `ackCount` bits and use success or failure reason code based on same bit position in `successMask`.

jfallows (changes requested):

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -505,56 +526,42 @@ private void onMqttSubscribe(
...
-                            decodeTraceId, beginEx.buffer(), beginEx.offset(), beginEx.sizeof());
+                        serverStream.doMqttBeginEx(decodeTraceId, topicFilter, subscriptionId);
 
                         correlations.put(newReplyId, serverStream);
 
-                        reasonCode = 0x00;
-
-                        mqttStreams.put(topicFilter, serverStream);
+                        subscribers.put(topicFilter, serverStream);
+                        reasonCodeCount++;
```
jfallows: This increments `reasonCodeCount` only when routeable, but we want the total count.
Can we just use `ackIndex` after the loop has completed? Perhaps define `ackIndex` outside the loop, but still increment it each iteration.
Perhaps rename `ackIndex` to `ackCount` since the count will be increasing, and remove `reasonCodeCount`.

jfallows: In fact, we could use `subscription.ackCount` directly instead.

jfallows (changes requested):

jfallows (dismissed):

jfallows (approved):

PR #20, opened 2019-10-21
Updating MqttServerFactory to be in line with the latest design in Http2ServerFactory

dknoma:
> Moving lambdas out from `MqttServer` to the factory and updated the functional interface to be in line with these new changes.

> Moving doFrame methods out to the factory as well as they are reused in the server, subscribe stream, and publish stream.

### DIFF
--- a/src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
+++ b/src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
@@ -143,13 +143,11 @@ public final class MqttServerFactory implements StreamFactory
 
     private final BufferPool bufferPool;
 
-    // private final MqttServerDecoder decodePreface = this::decodePreface;
     private final MqttServerDecoder decodePacketType = this::decodePacketType;
     private final MqttServerDecoder decodeConnect = this::decodeConnect;
     private final MqttServerDecoder decodePublish = this::decodePublish;
     private final MqttServerDecoder decodeSubscribe = this::decodeSubscribe;
     private final MqttServerDecoder decodeUnsubscribe = this::decodeUnsubscribe;
-    // private final MqttServerDecoder decodeUnsuback = this::decodeUnsuback;
     private final MqttServerDecoder decodePingreq = this::decodePingreq;
     private final MqttServerDecoder decodeDisconnect = this::decodeDisconnect;
     private final MqttServerDecoder decodeIgnoreOne = this::decodeIgnoreOne;
@@ -448,7 +446,6 @@ public final class MqttServerFactory implements StreamFactory
             final MqttPacketType packetType = MqttPacketType.valueOf(packet.typeAndFlags() >> 4);
             final MqttServerDecoder decoder = decodersByPacketType.getOrDefault(packetType, decodeIgnoreOne);
 
-            // if (packet.limit() + packet.remainingLength() <= limit)
             if (limit - packet.limit() >= length)
             {
                 server.decoder = decoder;
@@ -467,11 +464,7 @@ public final class MqttServerFactory implements StreamFactory
         final int limit)
     {
         MqttConnectFW mqttConnect = mqttConnectRO.wrap(buffer, offset, limit);
-        final boolean success = server.onMqttConnect(authorization, mqttConnect);
-        if (success)
-        {
-            server.decoder = decodePacketType;
-        }
+        server.onMqttConnect(authorization, mqttConnect);
         return mqttConnect.limit();
     }
 
@@ -520,23 +513,6 @@ public final class MqttServerFactory implements StreamFactory
         return unsubscribe.limit();
     }
 
-    // private int decodeUnsuback(
-    //     MqttServer server,
-    //     final long traceId,
-    //     final long authorization,
-    //     final long budgetId,
-    //     final DirectBuffer buffer,
-    //     final int offset,
-    //     final int limit)
-    // {
-    //     // server.doMqttUnsuback(authorization, );
-    //     // final MqttUnsubscribeFW unsubscribe = mqttUnsubscribeRO.wrap(buffer, offset, limit);
-    //     // server.onMqttUnsubscribe(authorization, unsubscribe);
-    //     // return unsubscribe.limit();
-    //     server.decoder = decodePacketType;
-    //     return offset;
-    // }
-
     private int decodePingreq(
         MqttServer server,
         final long traceId,
@@ -567,59 +543,6 @@ public final class MqttServerFactory implements StreamFactory
         server.doNetworkEnd(traceId);
         return disconnect.limit();
     }
-
-    // private int decodeSession(
-    //     MqttServer server,
-    //     final long traceId,
-    //     final long authorization,
-    //     final long budgetId,
-    //     final DirectBuffer buffer,
-    //     final int offset,
-    //     final int limit)
-    // {
-    //     int progress = 0;
-    //
-    //     final MqttPacketFixedHeaderFW packet = mqttPacketFixedHeaderRO.tryWrap(buffer, offset, limit);
-    //
-    //     if (packet != null && packet.limit() + packet.remainingLength() <= limit)
-    //     {
-    //         final MqttPacketType packetType = MqttPacketType.valueOf(packet.typeAndFlags() >> 4);
-    //         switch (packetType)
-    //         {
-    //             case PUBLISH:
-    //                 final MqttPublishFW publish = mqttPublishRO.wrap(buffer, offset, limit);
-    //                 server.onMqttPublish(authorization, publish);
-    //                 progress = publish.limit();
-    //                 break;
-    //             case SUBSCRIBE:
-    //                 final MqttSubscribeFW subscribe = mqttSubscribeRO.wrap(buffer, offset, limit);
-    //                 server.onMqttSubscribe(authorization, subscribe);
-    //                 progress = subscribe.limit();
-    //                 break;
-    //             case UNSUBSCRIBE:
-    //                 final MqttUnsubscribeFW unsubscribe = mqttUnsubscribeRO.wrap(buffer, offset, limit);
-    //                 server.onMqttUnsubscribe(authorization, unsubscribe);
-    //                 progress = unsubscribe.limit();
-    //                 break;
-    //             case PINGREQ:
-    //                 final MqttPingReqFW ping = mqttPingReqRO.wrap(buffer, offset, limit);
-    //                 server.onMqttPingReq(authorization, ping);
-    //                 progress = ping.limit();
-    //                 break;
-    //             case DISCONNECT:
-    //                 final MqttDisconnectFW disconnect = mqttDisconnectRO.wrap(buffer, offset, limit);
-    //                 server.onMqttDisconnect(disconnect);
-    //                 server.doNetworkEnd(traceId);
-    //                 progress = disconnect.limit();
-    //                 break;
-    //             default:
-    //                 server.doNetworkReset(traceId);
-    //                 break;
-    //         }
-    //     }
-    //
-    //     return progress;
-    // }
 
     private int decodeIgnoreOne(
         MqttServer server,
@@ -789,36 +712,6 @@ public final class MqttServerFactory implements StreamFactory
                 }
 
                 decodeNetwork(traceId, authorization, budgetId, buffer, offset, limit);
-
-                // while (offset < limit)
-                // {
-                //     // final MqttPacketFixedHeaderFW packet = mqttPacketFixedHeaderRO.tryWrap(buffer, offset, limit);
-                //     //
-                //     // if (packet == null || packet.limit() + packet.remainingLength() > limit)
-                //     // {
-                //     //     break;
-                //     // }
-                //     //
-                //     // final MqttPacketType packetType = MqttPacketType.valueOf(packet.typeAndFlags() >> 4);
-                //     offset += decoder.decode(this, authorization, decodeTraceId, buffer, offset, limit);
-                // }
-                //
-                // if (offset < limit)
-                // {
-                //     if (slotIndex == NO_SLOT)
-                //     {
-                //         slotIndex = bufferPool.acquire(streamId);
-                //     }
-                //     final MutableDirectBuffer slotBuffer = bufferPool.buffer(slotIndex);
-                //     slotLimit = limit - offset;
-                //     slotBuffer.putBytes(0, buffer, offset, slotLimit);
-                // }
-                // else if (slotIndex != NO_SLOT)
-                // {
-                //     bufferPool.release(slotIndex);
-                //     slotIndex = NO_SLOT;
-                //     slotLimit = 0;
-                // }
             }
         }
 
@@ -873,13 +766,13 @@ public final class MqttServerFactory implements StreamFactory
             doNetworkSignal(traceId);
         }
 
-        private boolean onMqttConnect(
+        private void onMqttConnect(
             long authorization,
             MqttConnectFW packet)
         {
-            int reasonCode = 0x00;
-
             final String protocolName = packet.protocolName().asString();
+
+            int reasonCode = 0x00;
             if (receivedConnect)
             {
                 reasonCode = 0x82; // protocol error
@@ -892,19 +785,17 @@ public final class MqttServerFactory implements StreamFactory
             {
                 reasonCode = 0x81; // malformed packet
             }
-            receivedConnect = true;
 
-            boolean success = reasonCode == 0;
-            if (success)
+            if (reasonCode == 0)
             {
                 doMqttConnack(authorization, reasonCode);
+                decoder = decodePacketType;
             }
             else
             {
                 doMqttConnack(authorization, reasonCode);
                 doNetworkEnd(decodeTraceId);
             }
-            return success;
         }
 
         private void onMqttPingReq(
@@ -1164,6 +1055,8 @@ public final class MqttServerFactory implements StreamFactory
                 .propertiesLength(properties.sizeof())
                 .properties(properties)
                 .build();
+
+            receivedConnect = true;
 
             doNetworkData(authorization, 0L, connack);
         }
--- a/src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
+++ b/src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
@@ -343,12 +343,16 @@ public final class MqttServerFactory implements StreamFactory
         MessageConsumer receiver,
         long routeId,
         long replyId,
-        long traceId)
+        long traceId,
+        long authorization,
+        Flyweight extension)
     {
         final EndFW end = endRW.wrap(writeBuffer, 0, writeBuffer.capacity())
                                .routeId(routeId)
                                .streamId(replyId)
                                .traceId(traceId)
+                               .authorization(authorization)
+                               .extension(extension.buffer(), extension.offset(), extension.sizeof())
                                .build();
 
         receiver.accept(end.typeId(), end.buffer(), end.offset(), end.sizeof());
@@ -476,14 +480,14 @@ public final class MqttServerFactory implements StreamFactory
 
         if (reasonCode == 0)
         {
-            server.onDecodeConnect(authorization, reasonCode, mqttConnect);
+            server.onDecodeConnect(traceId, authorization, reasonCode, mqttConnect);
             server.decoder = decodePacketType;
             progress = mqttConnect.limit();
         }
         else
         {
-            server.doEncodeConnack(authorization, reasonCode);
-            server.doNetworkEnd(traceId);
+            server.doEncodeConnack(traceId, authorization, reasonCode);
+            server.doNetworkEnd(traceId, authorization);
             server.decoder = decodeIgnoreAll;
         }
 
@@ -606,7 +610,7 @@ public final class MqttServerFactory implements StreamFactory
         }
         else
         {
-            server.onDecodeUnsubscribe(authorization, unsubscribe);
+            server.onDecodeUnsubscribe(traceId, authorization, unsubscribe);
             server.decoder = decodePacketType;
             progress = unsubscribe.limit();
         }
@@ -632,7 +636,7 @@ public final class MqttServerFactory implements StreamFactory
         }
         else
         {
-            server.onDecodePingReq(authorization, ping);
+            server.onDecodePingReq(traceId, authorization, ping);
             server.decoder = decodePacketType;
             progress = ping.limit();
         }
@@ -739,8 +743,15 @@ public final class MqttServerFactory implements StreamFactory
 
         private long decodeTraceId;
         private MqttServerDecoder decoder;
+
         private int decodeSlot = NO_SLOT;
         private int decodeSlotLimit;
+
+        private int encodeSlot = NO_SLOT;
+        private int encodeSlotOffset;
+        private long encodeSlotTraceId;
+        private int encodeSlotMaxLimit = Integer.MAX_VALUE;
+
         private boolean connected;
 
         private MqttServer(
@@ -861,7 +872,7 @@ public final class MqttServerFactory implements StreamFactory
                 cleanupDecodeSlotIfNecessary();
                 // cleanupSubscribersIfNecessary();
 
-                doNetworkEnd(traceId);
+                doNetworkEnd(traceId, authorization);
             }
             decoder = decodeIgnoreAll;
         }
@@ -903,18 +914,20 @@ public final class MqttServerFactory implements StreamFactory
         }
 
         private void onDecodeConnect(
+            long traceId,
             long authorization,
             int reasonCode,
             MqttConnectFW packet)
         {
-            doEncodeConnack(authorization, reasonCode);
+            doEncodeConnack(traceId, authorization, reasonCode);
         }
 
         private void onDecodePingReq(
+            long traceId,
             long authorization,
             MqttPingReqFW ping)
         {
-            doEncodePingResp(authorization);
+            doEncodePingResp(traceId, authorization);
         }
 
         private void onDecodeSubscribe(
@@ -988,6 +1001,7 @@ public final class MqttServerFactory implements StreamFactory
         }
 
         private void onDecodeUnsubscribe(
+            long traceId,
             long authorization,
             MqttUnsubscribeFW unsubscribe)
         {
@@ -1008,7 +1022,7 @@ public final class MqttServerFactory implements StreamFactory
                 topics++;
             }
 
-            doEncodeUnsuback(authorization, topics);
+            doEncodeUnsuback(traceId, authorization, topics);
         }
 
         private void onDecodePublish(
@@ -1047,7 +1061,7 @@ public final class MqttServerFactory implements StreamFactory
             MqttDisconnectFW disconnect)
         {
             /* process reason code */
-            doNetworkEnd(traceId);
+            doNetworkEnd(traceId, authorization);
         }
 
         private void onDecodeError(
@@ -1056,8 +1070,8 @@ public final class MqttServerFactory implements StreamFactory
             int reasonCode)
         {
             cleanupStreams();
-            doEncodeDisconnect(authorization, reasonCode);
-            doNetworkEnd(traceId);
+            doEncodeDisconnect(traceId, authorization, reasonCode);
+            doNetworkEnd(traceId, authorization);
         }
 
         private void doNetworkBegin(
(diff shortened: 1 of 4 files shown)


src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -159,7 +166,10 @@ public MqttServerFactory(
         this.supplyTraceId = requireNonNull(supplyTraceId);
         this.correlations = new Long2ObjectHashMap<>();
         this.wrapRoute = this::wrapRoute;
+        this.decodeConnectPacket = this::decodeConnectPacket;
```
jfallows: assign this inline at the field definition

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -159,7 +166,10 @@ public MqttServerFactory(
...
+        this.decodeSession = this::decodeSession;
```
jfallows: assign this inline at the field definition

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -159,7 +166,10 @@ public MqttServerFactory(
         this.supplyTraceId = requireNonNull(supplyTraceId);
         this.correlations = new Long2ObjectHashMap<>();
         this.wrapRoute = this::wrapRoute;
```
jfallows: Inline definition at field

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -159,7 +166,10 @@ public MqttServerFactory(
...
+        this.decodeConnectPacket = this::decodeConnectPacket;
+        this.decodeSession = this::decodeSession;
         this.mqttTypeId = supplyTypeId.applyAsInt(MqttNukleus.NAME);
+        this.connectionSet = new LongHashSet();
```
jfallows: This should not be needed.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -265,12 +280,223 @@ private RouteFW resolveTarget(
...
+            server.onMqttPublish(authorization, publish);
+            progress = publish.limit();
+            break;
+        default:
+            server.doNetworkReset(traceId);
+            break;
+        }
+
+        return progress;
+    }
```
jfallows, line 625: Change to `decodePacketType`, then based on observed `packetType`, change `decoder` to packet specific, such as `decodePing` or `decodeSubscribe`.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -265,12 +280,223 @@ private RouteFW resolveTarget(
...
+
+    @FunctionalInterface
+    private interface MqttServerDecoder
+    {
+        int decode(
+            MqttServer server,
+            long traceId,
+            long authorization,
+            MqttPacketType packetType,
```
jfallows: This parameter should not be needed in the decoder interface.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -282,22 +508,23 @@ private RouteFW resolveTarget(
...
+            long replyId,
+            long affinity)
         {
             this.network = network;
             this.routeId = routeId;
             this.initialId = initialId;
             this.replyId = replyId;
-            this.decodeState = this::decodeConnectPacket;
+            this.affinity = affinity;
+            this.decoder = decodeConnectPacket;
```
jfallows: `this.decoder = decodePacket`

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -437,58 +665,70 @@ private void onWindow(
...
             MqttConnectFW packet)
         {
             int reasonCode = 0x00;
-            // TODO: Must process multiple CONNECT from client as protocol errors and close network connection: 0x82
+            // TODO - Must process multiple CONNECT from client as protocol errors and close network connection: 0x82
+            //      - check via streamId/initialId
 
             final String protocolName = packet.protocolName().asString();
-            if (!"MQTT".equals(protocolName) || packet.protocolVersion() != 5)
+            if (connectionSet.contains(initialId))
```
jfallows: Just need to track current state of this `MqttServer`.

jfallows (changes requested):

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -265,12 +300,299 @@ private RouteFW resolveTarget(
...
+        final long traceId,
+        final long authorization,
+        final long budgetId,
+        final DirectBuffer buffer,
+        final int offset,
+        final int limit)
+    {
+        MqttConnectFW mqttConnect = mqttConnectRO.wrap(buffer, offset, limit);
+        server.onMqttConnect(authorization, mqttConnect);
+        return mqttConnect.limit();
```
jfallows: Use `tryWrap` instead of `wrap` to handle case where packet is incomplete.

Note that `tryWrap` will return `null` if it has insufficient bytes to successfully wrap.

This feedback applies to all the `decodeXXX` methods.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -265,12 +300,299 @@ private RouteFW resolveTarget(
...
+        MqttServer server,
+        final long traceId,
+        final long authorization,
+        final long budgetId,
+        final DirectBuffer buffer,
+        final int offset,
+        final int limit)
+    {
+        MqttConnectFW mqttConnect = mqttConnectRO.wrap(buffer, offset, limit);
+        server.onMqttConnect(authorization, mqttConnect);
```
jfallows: Rename `onMqttConnect` to `onDecodeConnect`, same feedback for other methods with similar naming structure.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -265,12 +300,299 @@ private RouteFW resolveTarget(
...
+        MqttServer server,
+        long traceId,
+        long authorization,
+        long budgetId,
+        DirectBuffer buffer,
+        int offset,
+        int limit)
+    {
+        final MqttPacketFixedHeaderFW packet = mqttPacketFixedHeaderRO.tryWrap(buffer, offset, limit);
+        final int progress = packet.limit();
```
jfallows: If `packet` does not include payload, then skipping would need to progress beyond `packet.limit()`.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -135,11 +138,43 @@
...
+    private final MqttServerDecoder decodePublish = this::decodePublish;
+    private final MqttServerDecoder decodeSubscribe = this::decodeSubscribe;
+    private final MqttServerDecoder decodeUnsubscribe = this::decodeUnsubscribe;
+    private final MqttServerDecoder decodePingreq = this::decodePingreq;
+    private final MqttServerDecoder decodeDisconnect = this::decodeDisconnect;
+    private final MqttServerDecoder decodeIgnoreOne = this::decodeIgnoreOne;
+    private final MqttServerDecoder decodeIgnoreAll = this::decodeIgnoreAll;
+
+
+    private final EnumMap<MqttPacketType, MqttServerDecoder> decodersByPacketType;
```
jfallows: Use type `Map<MqttPacketType, MqttServerDecoder>` for field.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -135,11 +138,43 @@
...
+    {
+        final EnumMap<MqttPacketType, MqttServerDecoder> decodersByPacketType = new EnumMap<>(MqttPacketType.class);
```
jfallows: Use type `Map<MqttPacketType, MqttServerDecoder>` for variable.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -135,11 +138,43 @@
...
+    private final MqttServerDecoder decodeSubscribe = this::decodeSubscribe;
+    private final MqttServerDecoder decodeUnsubscribe = this::decodeUnsubscribe;
+    private final MqttServerDecoder decodePingreq = this::decodePingreq;
+    private final MqttServerDecoder decodeDisconnect = this::decodeDisconnect;
+    private final MqttServerDecoder decodeIgnoreOne = this::decodeIgnoreOne;
+    private final MqttServerDecoder decodeIgnoreAll = this::decodeIgnoreAll;
+
+
+    private final EnumMap<MqttPacketType, MqttServerDecoder> decodersByPacketType;
+    {
```
jfallows, line 159: Add newline before this to show clear separation between member fields and non-static initializer block.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -640,104 +962,65 @@ private void onMqttDisconnect(
...
-                .streamId(initialId)
-                .traceId(traceId)
-                .build();
-
-            network.accept(signal.typeId(), signal.buffer(), signal.offset(), signal.sizeof());
+            doSignal(network, routeId, initialId, traceId);
         }
 
         private void doMqttPublish(
+            long authorization,
```
jfallows, line 1173: Rename `doMqttPublish` to `doEncodePublish`, same feedback for other methods with similar naming structure.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -265,12 +300,299 @@ private RouteFW resolveTarget(
...
+        MqttServer server,
+        final long traceId,
+        final long authorization,
+        final long budgetId,
+        final DirectBuffer buffer,
+        final int offset,
+        final int limit)
+    {
+        final MqttPublishFW publish = mqttPublishRO.wrap(buffer, offset, limit);
+        server.onMqttPublish(authorization, publish);
```
jfallows: Rename `onMqttPublish` to `onDecodePublish`

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -282,25 +604,28 @@ private RouteFW resolveTarget(
...
 
         private long decodeTraceId;
-        private DecoderState decodeState;
-        private int slotIndex = NO_SLOT;
-        private int slotLimit;
-        private long affinity;
+        private MqttServerDecoder decoder;
+        private int decodeSlot = NO_SLOT;
+        private int decodeSlotLimit;
+        private boolean receivedConnect;
```
jfallows: Rename `receivedConnect` to `connected`.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -265,12 +300,299 @@ private RouteFW resolveTarget(
...
+        final long authorization,
+        final long budgetId,
+        final DirectBuffer buffer,
+        final int offset,
+        final int limit)
+    {
+        final MqttDisconnectFW disconnect = mqttDisconnectRO.wrap(buffer, offset, limit);
+        server.onMqttDisconnect(disconnect);
+        server.decoder = decodePacketType;
+        server.doNetworkEnd(traceId);
```
jfallows: This line should be part of `onDecodeDisconnect` instead.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -265,12 +300,299 @@ private RouteFW resolveTarget(
...
+
+    private int decodeConnect(
+        MqttServer server,
+        final long traceId,
+        final long authorization,
+        final long budgetId,
+        final DirectBuffer buffer,
+        final int offset,
+        final int limit)
+    {
```
jfallows, line 468: Check protocol correctness in `decodeXXX` methods, potentially rejecting the frame and forcing cleanup if protocol is malformed in some way. For example, detecting a second connect when already connected should happen here.

But defer logical processing of correct protocol use to `server.onDecodeXXX` methods.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -503,8 +821,8 @@ private void onMqttSubscribe(
 
             if (limit == 0)
             {
-                doMqttDisconnect(0x82);
-                doEnd(decodeTraceId);
+                doMqttDisconnect(authorization, 0x82);
+                doNetworkEnd(decodeTraceId);
```
jfallows: These protocol-specific checks should be in `decodeSubscribe` instead.

jfallows (changes requested):

src/test/java/org/reaktivity/nukleus/mqtt/internal/ConnectionIT.java
```diff
@@ -117,6 +117,15 @@ public void shouldRejectMalformedConnectPacket() throws Exception
         k3po.finish();
     }
 
+    @Test
+    @Specification({
+        "${route}/server/controller",
+        "${client}/connect/reject.second.connect/client"})
+    public void shouldRejectDuplicateConnectPacket() throws Exception
```
jfallows: Update test name to reflect updated scenario name.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -135,11 +138,43 @@
     private final LongSupplier supplyTraceId;
 
     private final Long2ObjectHashMap<MqttServer.MqttServerStream> correlations;
-    private final MessageFunction<RouteFW> wrapRoute;
+    private final MessageFunction<RouteFW> wrapRoute = this::wrapRoute;
```
jfallows: Change to inline lambda definition instead of method reference.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -265,12 +300,375 @@ private RouteFW resolveTarget(
...
+        final long authorization,
+        final long budgetId,
+        final DirectBuffer buffer,
+        final int offset,
+        final int limit)
+    {
+        final MqttPingReqFW ping = mqttPingReqRO.tryWrap(buffer, offset, limit);
+        server.onDecodePingReq(authorization, ping);
+        server.decoder = decodePacketType;
+        return ping.limit();
```
jfallows: Handle `ping` packet same way as others.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -265,12 +300,375 @@ private RouteFW resolveTarget(
...
+
+        if (reasonCode == 0)
+        {
+            server.onDecodeConnect(authorization, reasonCode, mqttConnect);
+            server.decoder = decodePacketType;
+            progress = mqttConnect.limit();
+        }
+        else
+        {
+            server.onDecodeConnect(authorization, reasonCode, mqttConnect);
```
jfallows: If the decode fails with non-zero `reasonCode`, no need to call `onDecodeConnect`, just handle failure case here instead, then `onDecodeConnect` can be reserved for successful decode case only.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -265,12 +300,375 @@ private RouteFW resolveTarget(
...
+            server.doNetworkEnd(traceId);
+        }
+        else
+        {
+            server.onDecodeUnsubscribe(authorization, unsubscribe);
+            progress = unsubscribe.limit();
+        }
+
+        server.decoder = decodePacketType;
+        return progress;
```
jfallows, line 572: Apply equivalent feedback from `decodePublish` method.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -265,12 +300,375 @@ private RouteFW resolveTarget(
...
+        final DirectBuffer buffer,
+        final int offset,
+        final int limit)
+    {
+        final MqttPublishFW publish = mqttPublishRO.tryWrap(buffer, offset, limit);
+        int progress = limit;
+        if (limit == 0 || publish == null)
+        {
+            server.doEncodeDisconnect(authorization, 0x82);
+            server.doNetworkEnd(traceId);
```
jfallows: Refactor this as `onDecodeError` passing error code.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -265,12 +300,375 @@ private RouteFW resolveTarget(
...
+        final long traceId,
+        final long authorization,
+        final long budgetId,
+        final DirectBuffer buffer,
+        final int offset,
+        final int limit)
+    {
+        final MqttPublishFW publish = mqttPublishRO.tryWrap(buffer, offset, limit);
+        int progress = limit;
+        if (limit == 0 || publish == null)
```
jfallows: Remove `limit == 0` check, as it would imply `publish == null` anyway.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -265,12 +300,375 @@ private RouteFW resolveTarget(
...
+            server.doEncodeDisconnect(authorization, 0x82);
+            server.doNetworkEnd(traceId);
+        }
+        else
+        {
+            server.onDecodePublish(authorization, publish);
+            progress = publish.limit();
+        }
+
+        server.decoder = decodePacketType;
```
jfallows: Advance decoder in successful `else` clause only, switch to `decodeIgnoreAll` in failure `if` case.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -265,12 +300,375 @@ private RouteFW resolveTarget(
...
+        MqttServer server,
+        final long traceId,
+        final long authorization,
+        final long budgetId,
+        final DirectBuffer buffer,
+        final int offset,
+        final int limit)
+    {
+        final MqttPublishFW publish = mqttPublishRO.tryWrap(buffer, offset, limit);
+        int progress = limit;
```
jfallows: `progress` should initialize to `offset`, i.e. no progress yet

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -265,12 +300,375 @@ private RouteFW resolveTarget(
...
+            server.doNetworkEnd(traceId);
+        }
+        else
+        {
+            server.onDecodeSubscribe(authorization, subscribe);
+            progress = subscribe.limit();
+        }
+
+        server.decoder = decodePacketType;
+        return progress;
```
jfallows, line 546: Apply equivalent feedback from `decodePublish` method.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -265,12 +300,375 @@ private RouteFW resolveTarget(
...
+        }
+        else
+        {
+            server.onDecodeDisconnect(traceId, disconnect);
+            progress = disconnect.limit();
+        }
+
+        server.decoder = decodePacketType;
+        return progress;
+    }
```
jfallows, line 625: Apply equivalent feedback from `decodePublish` method.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -265,12 +300,375 @@ private RouteFW resolveTarget(
...
+    private int decodeIgnoreOne(
+        MqttServer server,
+        long traceId,
+        long authorization,
+        long budgetId,
+        DirectBuffer buffer,
+        int offset,
+        int limit)
+    {
+        final MqttPacketFixedHeaderFW packet = mqttPacketFixedHeaderRO.tryWrap(buffer, offset, limit);
```
jfallows: Use `wrap` instead of `tryWrap` because `decodePacketType` guarantees a successful wrap.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -265,12 +300,375 @@ private RouteFW resolveTarget(
...
+
+        int progress = limit;
+        if (packet != null)
+        {
+            progress = packet.limit();
+        }
+
+        server.decoder = decodePacketType;
+        return progress;
```
jfallows: Once switched to `wrap`, change above to:
```
server.decoder = decodePacketType;
return packet.limit();
```

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -265,12 +300,375 @@ private RouteFW resolveTarget(
...
+        {
+            server.onDecodeDisconnect(traceId, disconnect);
+            progress = disconnect.limit();
+        }
+
+        server.decoder = decodePacketType;
+        return progress;
+    }
+
+    private int decodeIgnoreOne(
```
jfallows: Does the MQTT specification have explicit language indicating that an unrecognized packet type should be skipped and processing continue on subsequent packets?

If not, then we should remove the `decodeIgnoreOne` method.

jfallows (changes requested):

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
dknoma: Self reminder:
```
in http2 we have onDecodeError trigger the cleanup for any existing application streams and also write out the GOAWAY frame, which is similar in spirit to mqtt DISCONNECT (edited) 
then after calling onDecodeError we set the decoder to decodeIgnoreAll
see doEncodeGoAway which for mqtt would be doEncodeDisconnect
new messages
then in doNetworkData after encode buffer is cleaned up, we check to see if no application streams are in flight and decoder is ignore-all, then we trigger doNetworkEnd
also in onNetworkEnd if decodeSlot is empty (so all pending frames have been processed then we cleanup any application streams in flight and doNetworkEnd
note: there is a TODO there to let the inbound complete normally and then let the outbound close on flush, but we don’t have that in place just yet and right now forcibly cleanup the application streams
this is typically a non-issue in practice, but we would prefer to make it better going forward, so recommend you keep the comment too as a reminder
```

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -135,11 +138,45 @@
     private final LongSupplier supplyTraceId;
 
     private final Long2ObjectHashMap<MqttServer.MqttServerStream> correlations;
-    private final MessageFunction<RouteFW> wrapRoute;
+    private final MessageFunction<RouteFW> wrapRoute = (msgTypeId, buffer, index, length) ->
+            routeRO.wrap(buffer, index, index + length);
```
jfallows: Use `(t, b, i, l)` for brevity instead of `(msgTypeId, buffer, index, length)` in lambda expression.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -262,7 +290,439 @@ private RouteFW resolveTarget(
...
+        final MqttPublishFW publish = mqttPublishRO.tryWrap(buffer, offset, limit);
+        int progress = offset;
+        if (publish == null)
+        {
+            server.onDecodeError(traceId, authorization, 0x82);
+            server.decoder = decodeIgnoreAll;
+        }
+        else
+        {
+            final String topicName = publish.topicName().asString();
```
jfallows: This logic on successful decode should be in `server.onDecodePublish(MqttPublishFW)` method instead.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -262,7 +290,439 @@ private RouteFW resolveTarget(
...
+        long authorization,
+        long budgetId,
+        DirectBuffer buffer,
+        int offset,
+        int limit)
+    {
+        final MqttPacketFixedHeaderFW packet = mqttPacketFixedHeaderRO.wrap(buffer, offset, limit);
+        server.decoder = decodePacketType;
+        return packet.limit();
+    }
```
jfallows: If this is unused, please remove it.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -262,7 +290,439 @@ private RouteFW resolveTarget(
...
+        long authorization,
+        long budgetId,
+        DirectBuffer buffer,
+        int offset,
+        int limit)
+    {
+        return limit;
+    }
+
+    private int decodeError(
```
jfallows: Rename to `decodeUnknownType`, same for matching lambda field.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -582,162 +1026,155 @@ private void onMqttUnsubscribe(
...
-                .correlationInfo(c -> c.bytes(b -> b.set(info.getBytes(UTF_8))))
-                .build();
-
-            OctetsFW payload = publish.payload();
-
-            MqttPublishStream publishStream = publishers.get(topicName);
-
-            final RouteFW route = resolveTarget(routeId, authorization, topicName);
-            if (route != null)
+            if (publishStream != null)
```
jfallows: `MqttPublishStream` should ideally only be used inside `MqttServer` not in shared `decodePublish` method on factory.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -262,7 +290,439 @@ private RouteFW resolveTarget(
...
+        if (reasonCode == 0)
+        {
+            server.onDecodeConnect(traceId, authorization, reasonCode, mqttConnect);
+            server.decoder = decodePacketType;
+            progress = mqttConnect.limit();
+        }
+        else
+        {
+            server.doEncodeConnack(traceId, authorization, reasonCode);
+            server.doNetworkEnd(traceId, authorization);
```
jfallows: Use `server.onDecodeError` here too, then use `connected` to determine whether `onDecodeError` should `doEncodeConnack` or `doEncodeDisconnect`.

jfallows (changes requested):

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -616,128 +997,147 @@ private void onMqttPublish(
...
+            boolean isConnectError)
         {
-            final BeginFW begin = beginRW.wrap(writeBuffer, 0, writeBuffer.capacity())
-                .routeId(routeId)
-                .streamId(replyId)
-                .traceId(traceId)
-                .affinity(affinity)
-                .build();
+            cleanupStreams();
+            if (isConnectError)
```
jfallows: Can we use `connected` field here instead and remove `isConnectError ` parameter?

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -582,30 +962,31 @@ private void onMqttUnsubscribe(
                 topics++;
             }
 
-            doMqttUnsuback(topics);
+            doEncodeUnsuback(traceId, authorization, topics);
         }
 
-        private void onMqttPublish(
+        private void onDecodePublish(
+            long authorization,
```
jfallows, line 1056: We'll need `traceId` here too, like other `onDecodeXXX` methods.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -616,128 +997,147 @@ private void onMqttPublish(
                 final MessageConsumer newTarget = router.supplyReceiver(newInitialId);
                 if (publishStream != null)
                 {
-                    publishStream.doMqttDataEx(decodeTraceId, payload, dataEx);
+                    publishStream.doMqttDataEx(decodeTraceId, authorization, payload, dataEx);
```
jfallows: Use passed in `traceId` instead - see comment above.

jfallows (changes requested):

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -271,36 +669,49 @@ private RouteFW resolveTarget(
...
             this.initialId = initialId;
             this.replyId = replyId;
-            this.decodeState = this::decodeConnectPacket;
+            this.affinity = affinity;
+            this.budgetId = budgetId;
+            this.decoder = decodePacketType;
             this.subscribers = new HashMap<>();
             this.publishers = new HashMap<>();
             this.subscriptionsByPacketId = new Int2ObjectHashMap<>();
+            this.connected = false;
```
jfallows: No need for explicit assignment to language defined default - this line should be removed.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -785,20 +1176,26 @@ private void doMqttConnack(
                 .properties(properties)
                 .build();
 
-            doData(connack);
+            connected = true;
```
jfallows: This should be set in `onDecodeConnect` instead since we are not really `connected` when sending error on failed initial connect.

jfallows (changes requested):

jfallows (dismissed):

jfallows (approved):

PR #22, opened 2019-11-19
PUBLISH to topic name

dknoma:


src/test/java/org/reaktivity/nukleus/mqtt/internal/ConnectionIT.java
```diff
@@ -110,6 +110,18 @@ public void shouldExchangeConnectionPacketsThenPublishMultipleMessages() throws
         k3po.finish();
     }
 
+    @Test
+    @Specification({
+        "${route}/server/controller",
+        "${client}/publish/send.multiple.messages/client",
+        "${server}/send.multiple.messages.then.timeout/server"})
+    public void shouldExchangeConnectionPacketsThenPublishMultipleMessagesThenTimeout() throws Exception
```
jfallows: Rename to `shouldPublishMultipleMessagesThenClose()`

src/test/java/org/reaktivity/nukleus/mqtt/internal/ConnectionIT.java
```diff
@@ -110,6 +110,18 @@ public void shouldExchangeConnectionPacketsThenPublishMultipleMessages() throws
...
+    {
+        k3po.start();
+        Thread.sleep(5000); // third message doesn't get delivered if publish stream timesout
```
jfallows: The behavior described by this comment is not what we are trying to model.

If there is a 3rd message that arrives after the timeout, then that will cause a new application publish stream to be created (as there is no longer an existing application publish stream to reuse) then the message should still be sent on the new application publish stream.

jfallows (changes requested):

jfallows (dismissed):

jfallows (approved):

PR #23, opened 2019-12-02
PUBLISH from topic name

### LINKED ISSUE #18: PUBLISH from topic name
When `SUBSCRIBED` to a topic, `DATA` frames arriving on a `reaktivity` stream should be delivered to the `mqtt` subscriber as `PUBLISH` frames.

dknoma:


src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -749,6 +751,8 @@ int decode(
         private final long affinity;
         private final long budgetId;
 
+        private MqttSettings remoteSettings;
+
```
jfallows: This concept of settings comes from `http/2` which negotiates client and server capabilities during the initial handshake.
In `mqtt` there doesn't seem to be an equivalent, so we probably don't need to model this settings concept.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -906,14 +913,31 @@ private void onNetworkAbort(
...
-            replyBudget += replyCredit;
-            replyPadding += window.padding();
+                encodeNetwork(encodeSlotTraceId, authorization, budgetId, buffer, 0, limit, maxLimit);
+            }
 
-            final int initialCredit = bufferPool.slotCapacity() - initialBudget;
+            if (encodeSlot == NO_SLOT)
+            {
+                subscribers.values().forEach(sub -> sub.flushReplyWindow(traceId, authorization));
+                publishers.values().forEach(pub -> pub.flushReplyWindow(traceId, authorization));
```
jfallows: Publishers are not expecting any `DATA` frames, so no need to flush the reply window for publishers.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -995,6 +1019,8 @@ private void onDecodePublish(
                 });
                 publishStream.doApplicationData(traceId, authorization, payload, dataEx);
                 correlations.put(newReplyId, publishStream);    // TODO: do we need to clean up correlations onAbort()?
+
+                // doEncodePublish(traceId, authorization, topicName, );
```
jfallows: We only need to `doEncodePublish` in reaction to an application `DATA` frame arriving that then needs to be sent to the `mqtt` client, so please remove this commented line.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1052,6 +1078,7 @@ private void onDecodeSubscribe(
                                                     newRouteId, newInitialId, newReplyId, packetId, topicFilter, subscription);
 
                     subscribeStream.doApplicationBegin(traceId, authorization, affinity, topicFilter, subscriptionId);
+                    subscribeStream.doApplicationData(traceId, authorization, buffer, offset, limit);
```
jfallows: We shouldn't be sending `DATA` frames towards the application for a subscriber.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1222,9 +1250,8 @@ private void doEncodePublish(
             String topicName,
             OctetsFW payload)
         {
-            OctetsFW properties = octetsRW
-                                      .wrap(writeBuffer, 0, 0)
-                                      .build();
+            OctetsFW properties = octetsRW.wrap(writeBuffer, 0, 0)
+                                          .build();
```
jfallows: Pre-create a shared empty `OctetsFW` instead of wrapping on every publish.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1242,9 +1269,8 @@ private void doEncodeConnack(
             long authorization,
             int reasonCode)
         {
-            OctetsFW properties = octetsRW
-                                      .wrap(writeBuffer, 0, 0)
-                                      .build();
+            OctetsFW properties = octetsRW.wrap(writeBuffer, 0, 0)
+                                          .build();
```
jfallows: Use shared empty `OctetsFW` instead.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1571,6 +1597,8 @@ abstract void onApplicationReply(
             private int initialPadding;
             private int replyBudget;
 
+            private int remoteBudget;
+
```
jfallows: Remote budget is an `http/2` concept, so we don't need to model it for `mqtt`.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1796,23 +1828,24 @@ public void onApplicationReply(
             private void onApplicationBegin(
                 BeginFW begin)
             {
+                state = MqttState.openReply(state);
+
+                final long traceId = begin.traceId();
+                final long authorization = begin.authorization();
+
+                onReplyWindowUpdate(traceId, authorization, remoteSettings.initialWindowSize);
```
jfallows: No need to model response window, this should just send non-zero window for subscribe stream, but zero window for publish stream, as we don't expect any `DATA` frames to be sent back on the publish stream.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1796,23 +1828,24 @@ public void onApplicationReply(
...
             }
 
             private void onApplicationData(
                 DataFW data)
             {
                 final long traceId = data.traceId();
```
jfallows: Check the reply budget and `RESET` if it is exceeded, then cleanup.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1821,6 +1854,45 @@ private void onApplicationEnd(
...
+                    if (responseCredit > 0)
+                    {
+                        replyBudget += responseCredit;
+
+                        doWindow(application, routeId, replyId, traceId, authorization,
+                            budgetId, responseCredit, 0);
+                    }
+                }
+            }
+
```
jfallows, line 1848: This is a complex flow control concept brought over from `http/2` that we don't need to model in `mqtt`.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -2074,13 +2146,30 @@ private void onApplicationWindow(
                 initialBudget += credit;
                 initialPadding = padding;
 
-                subscription.onSubscribeSucceeded(traceId, authorization, packetId, ackIndex);
+                if (!isSubscribed)
+                {
+                    isSubscribed = true;
+                    subscription.onSubscribeSucceeded(traceId, authorization, packetId, ackIndex);
+                }
```
jfallows, line 2104: Use `state` instead of `isSubscribed`.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -2074,13 +2146,30 @@ private void onApplicationWindow(
...
+                    final int limit = initialSlotOffset;
+
+                    flushApplicationData(initialSlotTraceId, authorization, buffer, offset, limit);
+                }
 
-                if (!MqttState.initialClosed(padding))
+                if (initialSlot == NO_SLOT)
                 {
-                    if (MqttState.initialClosing(state))
+                    if (!MqttState.initialClosed(padding))
```
jfallows: If `initial` is `closed` then we must have already sent an `ABORT`, `END` or received `RESET`, so then the `reaktor` will not call us with `WINDOW`, so no need to defend this case?

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -2126,7 +2220,10 @@ private void onApplicationBegin(
             {
                 state = MqttState.openReply(state);
 
-                final MqttBeginExFW beginEx = begin.extension().get(mqttBeginExRO::tryWrap);
+                final long traceId = begin.traceId();
+                final long authorization = begin.authorization();
+
+                onReplyWindowUpdate(traceId, authorization, remoteSettings.initialWindowSize);
```
jfallows: See above comments about flow control.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -2169,6 +2266,93 @@ private void doApplicationResetIfNecessary(
...
+                        initialSlotOffset = remaining;
+                        initialSlotTraceId = traceId;
+                    }
+                }
+                else
+                {
+                    cleanupInitialSlotIfNecessary();
+                }
+            }
+
```
jfallows, line 2293: See above comments about flow control.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttSettings.java
```diff
@@ -0,0 +1,56 @@
...
+    private boolean enablePushIsValid()
+    {
+        return enablePush == 0 || enablePush == 1;
+    }
+
+    private boolean initialWindowSizeIsValid()
+    {
+        return initialWindowSize >= 0;
+    }
+}
```
jfallows: We probably don't need to model this in `mqtt`.

jfallows (changes requested):

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -906,14 +908,30 @@ private void onNetworkAbort(
...
+                final int limit = Math.min(encodeSlotOffset, encodeSlotMaxLimit);
+                final int maxLimit = encodeSlotOffset;
 
-            final int initialCredit = bufferPool.slotCapacity() - initialBudget;
+                encodeNetwork(encodeSlotTraceId, authorization, budgetId, buffer, 0, limit, maxLimit);
+            }
+
+            if (encodeSlot == NO_SLOT)
+            {
+                subscribers.values().forEach(sub -> sub.flushReplyWindow(traceId, authorization));
```
jfallows: Use single letter `s` rather than `sub` for lambda expressions.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1092,7 +1110,7 @@ private void onDecodeUnsubscribe(
                 {
                     break;
                 }
-                final int topicKey = System.identityHashCode(topic.filter().asString().intern());
+                final int topicKey = topicKey(topic.filter().asString());
```
jfallows, line 1113: Does `UNSUBSCRIBE` need a protocol check for `topicFilter == null`, like `SUBSCRIBE`?

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1061,7 +1079,7 @@ private void onDecodeSubscribe(
                         return;
```
jfallows: Try to avoid multi-return methods, instead break the loop and let the default return at end of method take effect.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1327,17 +1338,13 @@ private void doEncodeDisconnect(
...
-
             final MqttDisconnectFW disconnect = mqttDisconnectRW
                                                     .wrap(writeBuffer, DataFW.FIELD_OFFSET_PAYLOAD, writeBuffer.capacity())
                                                     .typeAndFlags(0xe0)
-                                                    .remainingLength((byte) properties.sizeof() + 2)
+                                                    .remainingLength((byte) EMPTY_OCTETS.sizeof() + 2)
                                                     .reasonCode(reasonCode)
-                                                    .propertiesLength(properties.sizeof())
-                                                    .properties(properties)
+                                                    .propertiesLength(EMPTY_OCTETS.sizeof())
```
jfallows: Is `propertiesLength` not computed automatically when `properties` is called?

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1971,18 +1995,20 @@ private void doApplicationBegin(
                 doBegin(application, routeId, initialId, traceId, authorization, affinity, beginEx);
                 final MutableInteger value = activeSubscribers.computeIfAbsent(topicFilter, key -> new MutableInteger());
```
jfallows: Suggest using `Int2ObjectHashMap` for active subscribers (and publishers), using `topicKey` approach to get predictable `int` keys for the map.
Note: `HashMap` is functionally correct, but produces more garbage.

jfallows (changes requested):

jfallows (approved):

PR #24, opened 2019-12-13
Unified publish subscribe stream

dknoma:


src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1088,12 +1107,16 @@ private void onDecodeUnsubscribe(
...
+                    onDecodeError(traceId, authorization, PROTOCOL_ERROR);
+                    return;
                 }
-                final int topicKey = System.identityHashCode(topic.filter().asString().intern());
-                subscribers.remove(topicKey);
+                final String topicString = topic.filter().asString();
+                final int topicKey = topicKey(topicString);
+                final MqttServerStream stream = streams.get(topicKey);
+                stream.roles.remove(MqttRole.RECEIVER);
+                // TODO - END to application; unsubscribe from topic, and publish timesout with end either way
```
jfallows: Still TODO ?

jfallows (changes requested):

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -981,16 +979,15 @@ private void onDecodePublish(
...
-                                            .build();
+            final MqttDataExFW dataEx = mqttDataExRW.wrap(dataExtBuffer, 0, dataExtBuffer.capacity())
+                                                    .typeId(mqttTypeId)
+                                                    .topic(topicName)
+                                                    .expiryInterval(15)
+                                                    .contentType("message")
+                                                    .format(f -> f.set(MqttPayloadFormat.TEXT))
+                                                    .responseTopic(topicName)
+                                                    .correlationInfo(c -> c.bytes(b -> b.set(info.getBytes(UTF_8))))
+                                                    .build();
```
jfallows, line 1050: Some of these hardcoded fields should be coming from `MqttPublishFW` frame instead, right?

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1003,15 +1000,15 @@ private void onDecodePublish(
...
+                MqttServerStream stream = streams.computeIfAbsent(topicKey, s ->
                 {
-                    final MqttPublishStream newPublishStream = new MqttPublishStream(newTarget,
-                                                                     newRouteId, newInitialId, newReplyId, 0, topicName);
-                    newPublishStream.doApplicationBegin(traceId, authorization, affinity);
-                    return newPublishStream;
+                    final MqttServerStream newStream = new MqttServerStream(newTarget, newRouteId, newInitialId, newReplyId,
+                                                        0, topicName, null, MqttRole.SENDER);
+                    newStream.doApplicationBegin(traceId, authorization, affinity, topicName, 0);
+                    return newStream;
```
jfallows: Suggest avoiding side-effects like `doApplicationBegin` during construction.
Instead, keep the `computeIfAbsent` mapping function to construction only, and then do a `stream.doApplicationBeginIfNecessary()` after wards that checks the application `state` to determine if `BEGIN` needs to be sent.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1003,15 +1000,15 @@ private void onDecodePublish(
...
                 });
-                publishStream.doApplicationData(traceId, authorization, payload, dataEx);
-                correlations.put(newReplyId, publishStream);    // TODO: do we need to clean up correlations onAbort()?
+                stream.doApplicationData(traceId, authorization, MqttRole.SENDER, payload, dataEx);
+                correlations.put(newReplyId, stream);    // TODO: do we need to clean up correlations onAbort()?
```
jfallows: Yes, correlations need to be tidied up for each application stream.
This can be handled when the application stream is asked to cleanup.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1065,21 +1062,26 @@ private void onDecodeSubscribe(
...
+
+                    int finalSubscriptionId = subscriptionId;
+                    MqttServerStream stream = streams.computeIfAbsent(topicKey, s ->
+                    {
+                        final MqttServerStream newStream = new MqttServerStream(newTarget,
+                            newRouteId, newInitialId, newReplyId, packetId, topicFilter, subscription,
+                            MqttRole.RECEIVER);
+                        newStream.doApplicationBegin(traceId, authorization, affinity, topicFilter, finalSubscriptionId);
+                        return newStream;
+                    });
```
jfallows: Same feedback as above regarding side effects during `computeIfAbsent` and using `doApplicationBeginIfNecessary` instead.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1110,8 +1112,15 @@ private void onDecodeUnsubscribe(
                     onDecodeError(traceId, authorization, PROTOCOL_ERROR);
                     return;
                 }
```
jfallows, line 1176: Suggest tracking the decode error in the loop and breaking the loop on problem.
Then after loop, check for error and call `onDecodeError`, otherwise call `doEncodeUnsubAck`.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1576,63 +1569,101 @@ abstract void onApplicationReply(
...
+            MqttServerStream(
                 MessageConsumer application,
                 long routeId,
                 long initialId,
                 long replyId,
                 int packetId,
-                String topicFilter)
+                String topicFilter,
+                Subscription subscription,
+                MqttRole role)
```
jfallows: Suggest moving the `subscription` specific stuff to a method and out of the constructor because the same stream might be used first for publishing.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1576,63 +1569,101 @@ abstract void onApplicationReply(
...
             {
                 assert state == 0;
                 state = MqttState.openingInitial(state);
 
-                this.signalFuture = executor.schedule(publishTimeout, SECONDS, routeId, initialId,
-                                                      PUBLISH_TIMEOUT_SIGNAL);
+                final MqttRole role;
+                if (this.roles.contains(MqttRole.SENDER))
+                {
+                    this.signalFuture = executor.schedule(publishTimeout, SECONDS, routeId, initialId,
```
jfallows: Change `mqtt` server to use `Signaler` instead of `SignalingExecutor` to avoid garbage creation of futures.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1576,63 +1569,101 @@ abstract void onApplicationReply(
...
+                }
 
                 router.setThrottle(initialId, this::onApplicationInitial);
 
                 final MqttBeginExFW beginEx = mqttBeginExRW
                                                   .wrap(extBuffer, 0, extBuffer.capacity())
                                                   .typeId(mqttTypeId)
-                                                  .role(r -> r.set(MqttRole.SENDER))
+                                                  .role(r -> r.set(role))
                                                   .clientId("client")
```
jfallows: This `"client"` value should come from configuration, with a suitable default.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1731,13 +1755,53 @@ public void onApplicationInitial(
...
+                    final MutableDirectBuffer buffer = bufferPool.buffer(initialSlot);
+                    final int offset = 0;
+                    final int limit = initialSlotOffset;
+
+                    flushApplicationData(initialSlotTraceId, authorization, buffer, offset, limit, EMPTY_OCTETS);
+                }
+
+                if (initialSlot == NO_SLOT)
+                {
+                    if (MqttState.initialClosing(state))
```
jfallows: `&& !MqttState.initialClosed(state)`

jfallows (changes requested):

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -977,20 +989,63 @@ private void onDecodePublish(
...
+                case KIND_TOPIC_ALIAS:
+                    break;
+                default:
+                    onDecodeError(traceId, authorization, MALFORMED_PACKET);
+                    break;
+                }
+            }
+
+            final MqttPayloadFormat format = mqttPayloadFormat != null ? mqttPayloadFormat : MqttPayloadFormat.TEXT;
+            final OctetsFW finalCorrelationInfo = correlationInfo;
```
jfallows: In these cases, where we need to make a non-final variable final for the sake of a lambda, we typically just use the same variable name and add a `0` at the end, so `mqttPayloadFormat0` and `correlationInfo0`.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1003,15 +1058,12 @@ private void onDecodePublish(
...
-                    final MqttPublishStream newPublishStream = new MqttPublishStream(newTarget,
-                                                                     newRouteId, newInitialId, newReplyId, 0, topicName);
-                    newPublishStream.doApplicationBegin(traceId, authorization, affinity);
-                    return newPublishStream;
-                });
-                publishStream.doApplicationData(traceId, authorization, payload, dataEx);
-                correlations.put(newReplyId, publishStream);    // TODO: do we need to clean up correlations onAbort()?
+                MqttServerStream stream = streams.computeIfAbsent(topicKey, s ->
+                                                        new MqttServerStream(newTarget, newRouteId, newInitialId, newReplyId,
+                                                            0, topicName, MqttRole.SENDER));
```
jfallows: Please move `newInitialId`, `newTarget` and `newReplyId` derivations into `MqttServerStream` constructor - only requires `newRouteId` to derive those.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1042,52 +1094,62 @@ private void onDecodeSubscribe(
...
+            if (subscriptionId == 0)
             {
-                mqttSubscription = mqttSubscriptionRO.tryWrap(buffer, progress, limit);
-                if (mqttSubscription == null)
-                {
-                    break;
-                }
-                final String topicFilter = mqttSubscription.topicFilter().asString();
-                final RouteFW route = resolveTarget(routeId, authorization, topicFilter);
+                onDecodeError(traceId, authorization, PROTOCOL_ERROR);
```
jfallows, line 1106: Please add an IT to verify that `subscriptionId` of `0` is rejected, per spec.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1042,52 +1094,62 @@ private void onDecodeSubscribe(
...
+            }
+            else
+            {
+                subscription.subscriptionId = subscriptionId;
```
jfallows: This is probably just `subscription.id`, right?

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1102,18 +1164,34 @@ private void onDecodeUnsubscribe(
             final int packetId = unsubscribe.packetId();
 
             MqttTopicFW topic;
+            boolean error = false;
```
jfallows: Retain the `int` error code instead of a `boolean`, initialized to `NO_ERROR` then becomes `PROTOCOL_ERROR`.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1102,18 +1164,34 @@ private void onDecodeUnsubscribe(
...
-                subscribers.remove(topicKey);
+                final String topicString = topic.filter().asString();
+                final int topicKey = topicKey(topicString);
+                final MqttServerStream stream = streams.get(topicKey);
+
+                final MqttEndExFW endEx = mqttEndExRW.wrap(extBuffer, 0, extBuffer.capacity())
+                                              .typeId(mqttTypeId)
+                                              .build();
+
+                stream.doApplicationEnd(traceId, authorization, endEx);
```
jfallows: What does the spec say about partially processing partially invalid `UNSUBSCRIBE` packets?
Is it all-or-nothing? If so then you'll need to do two passes over the topic filters to verify they are all valid before processing any of them.
Though if there is a decode error, then I suppose we are going to clean up everything anyway, so probably it is a moot point, agreed?

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1238,14 +1316,39 @@ private void doNetworkSignal(
...
+                .messageExpiryInterval(dataEx.expiryInterval())
+                .wrap(mqttPropertyBuffer, mqttPropertyRW.limit(), mqttPropertyBuffer.capacity())
+                .contentType(dataEx.contentType().asString())
+                .wrap(mqttPropertyBuffer, mqttPropertyRW.limit(), mqttPropertyBuffer.capacity())
+                .payloadFormatIndicator((byte) dataEx.format().get().ordinal())
+                .wrap(mqttPropertyBuffer, mqttPropertyRW.limit(), mqttPropertyBuffer.capacity())
+                .responseTopic(dataEx.responseTopic().asString())
+                .wrap(mqttPropertyBuffer, mqttPropertyRW.limit(), mqttPropertyBuffer.capacity())
+                .correlationData(a -> a.bytes(dataEx.correlationInfo().bytes()))
+                .build();
```
jfallows, line 1343: This is a little difficult to read, though logically correct.

Please split each property into a separate wrap + set + build to improve readability.

```
MqttPropertyFW.Builder mqttProperty = mqttPropertyRW.wrap(mqttPropertyBuffer, 0, mqttPropertyBuffer.capacity());

if (subscriptionId > 0)
{
    mqttProperty.wrap(mqttPropertyBuffer, mqttProperty.limit(), mqttPropertyBuffer.capacity())
                .subscriptionId(v -> v.set(subscriptionId));
}
mqttProperty.wrap(mqttPropertyBuffer, mqttProperty.limit(), mqttPropertyBuffer.capacity())
            .messageExpiryInterval(dataEx.expiryInterval())
            .build()
mqttProperty.wrap(mqttPropertyBuffer, mqttProperty.limit(), mqttPropertyBuffer.capacity())
            .contentType(dataEx.contentType().asString())
            .build()
...
```
Also, check if it is necessary to call `asString()`, would be better to avoid `UTF-8` decoding and encoding if not needed.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1238,14 +1316,39 @@ private void doNetworkSignal(
...
+
+            final int limit = mqttPropertyRW.limit();
+
             final MqttPublishFW publish = mqttPublishRW.wrap(writeBuffer, DataFW.FIELD_OFFSET_PAYLOAD, writeBuffer.capacity())
                                               .typeAndFlags(0x30)
                                               .remainingLength(0x14)
                                               .topicName(topicName)
-                                              .properties(EMPTY_OCTETS)
+                                              .propertiesLength(limit)
```
jfallows, line 1351: Is `propertiesLength` not automatic?

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1828,20 +2009,93 @@ private void onApplicationEnd(
...
+                    }
+
+                    if (initialSlot == NO_SLOT)
+                    {
+                        cleanup(traceId, authorization);
+                    }
+                    else
+                    {
+                        final MutableDirectBuffer requestBuffer = bufferPool.buffer(initialSlot);
+                        requestBuffer.putBytes(0, buffer, offset, remaining);
```
jfallows: `requestBuffer` -> `initialBuffer` ?

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1566,73 +1654,127 @@ abstract void onApplicationReply(
...
-                refreshPublishTimeout();
-
-                doData(application, routeId, initialId, traceId, authorization, 0L, payload.sizeof(),
-                       payload.buffer(), payload.offset(), payload.sizeof(), extension);
+                switch (role)
+                {
+                case SENDER:
+                    refreshPublishTimeout();
+                    doData(application, routeId, initialId, traceId, authorization, 0L, payload.sizeof(),
+                        payload.buffer(), payload.offset(), payload.sizeof(), extension);
```
jfallows: What if we don't have enough `WINDOW` on the application initial stream?

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1566,73 +1654,127 @@ abstract void onApplicationReply(
...
+                    break;
+                case RECEIVER:
+                default:
+                    break;
+                }
+                roles.add(role);
```
jfallows: Suggest we avoid this side effect and instead assert that we have `SENDER` role to be able to send (publish) `DATA` frames.

jfallows (changes requested):

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
dknoma: Yeah, agreed. Either way should honestly work the same.

jfallows (approved):

PR #26, opened 2020-01-09
Proper Wildcard Topic Filter handling

### LINKED ISSUE #13: SUBSCRIBE to single wildcard topic name pattern
Support MQTT subscription with a single topic filter consisting of a wildcard topic name pattern only.

dknoma:
Fix #13


src/main/java/org/reaktivity/nukleus/mqtt/internal/MqttUtility.java
```diff
@@ -0,0 +1,33 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package org.reaktivity.nukleus.mqtt.internal;
+
+import java.util.regex.Pattern;
+
+public final class MqttUtility
```
jfallows: Rename to `MqttUtil`.

src/main/java/org/reaktivity/nukleus/mqtt/internal/MqttUtility.java
```diff
@@ -0,0 +1,33 @@
...
+{
+    public static final Pattern TOPIC_FILTER_REGEX = Pattern.compile("[/]?(([^/#+]*|\\+)/)*(#|\\+|[^/#+]*)");
```
jfallows: Make constant `private`.

src/main/java/org/reaktivity/nukleus/mqtt/internal/MqttUtility.java
```diff
@@ -0,0 +1,33 @@
...
+
+    public static boolean validTopicFilter(String topicFilter)
```
jfallows: Rename to `isTopicFilterValid(topicFilter)`, also format param on next line.

src/main/java/org/reaktivity/nukleus/mqtt/internal/MqttUtility.java
```diff
@@ -0,0 +1,33 @@
...
+    {
+        return TOPIC_FILTER_REGEX.matcher(topicFilter).matches();
```
jfallows: Hoist matcher as thread local constant.
```
private static final ThreadLocal<Matcher> TOPIC_FILTER_MATCHER = ThreadLocal.withInitial(() -> TOPIC_FILTER_REGEX.matcher(""));
```
Then in this method, use:
```
    return TOPIC_FILTER_MATCHER.get().reset(topicFilter).matches();
```

src/main/java/org/reaktivity/nukleus/mqtt/internal/MqttUtility.java
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    private MqttUtility()
+    {
+        // Utility
+    }
+}
```
jfallows: Alternatively, rename this class to `MqttValidator`, make the `isTopicFilterValid(topicFilter)` method non-static, eliminate the need for `ThreadLocal`, but use the same approach to reuse the `Matcher` and create an instance of `MqttValidator` in `MqttServerFactory` that is reused by all `mqtt` streams on the same thread (a different thread will have its own `MqttServerFactory`).

src/test/java/org/reaktivity/nukleus/mqtt/internal/UtilityTest.java
```diff
@@ -0,0 +1,60 @@
...
+        }
+        else
+        {
+            for (String s : filters)
+            {
+                assertFalse(String.format("\"%s\" was a valid filter when it's supposed to be invalid.", s), validTopicFilter(s));
+            }
+        }
+    }
+}
```
jfallows: Instead of just two test methods, one for valid and one for invalid, separate each case into a separate method, such as `shouldValidateMultiLevelWildcard()`, `shouldValidateMultiLevelWildcardAtRoot()`, `shouldValidateMultiLevelWildcardAtEnd()`, `shouldValidateMultiLevelWildcardAfterSingleLevelWildcards()`, `shouldNotValidateMultiLevelWildcardBeforeSeparator()`, etc.

Then the method name becomes the description of what is actually being tested, and the code is much more self-documenting.

jfallows (changes requested):

src/main/java/org/reaktivity/nukleus/mqtt/internal/MqttValidator.java
```diff
@@ -0,0 +1,36 @@
...
+package org.reaktivity.nukleus.mqtt.internal;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public final class MqttValidator
+{
+    private static final Pattern TOPIC_FILTER_REGEX = Pattern.compile("[/]?(([^/#+]*|\\+)/)*(#|\\+|[^/#+]*)");
+
+    private static final Matcher TOPIC_FILTER_MATCHER = TOPIC_FILTER_REGEX.matcher("");
```
jfallows: This needs to be non-static because it is not thread-safe.

src/test/java/org/reaktivity/nukleus/mqtt/internal/ValidatorTest.java
```diff
@@ -0,0 +1,305 @@
...
+        assertTrue(String.format("\"%s\" was an invalid filter when it's supposed to be valid.", topicFilter),
+            validator.isTopicFilterValid(topicFilter));
+    }
+
+    private void checkForInvalidTopicFilter(
+        String topicFilter)
+    {
+        assertFalse(String.format("\"%s\" was a valid filter when it's supposed to be invalid.", topicFilter),
+            validator.isTopicFilterValid(topicFilter));
+    }
```
jfallows, line 223: Now that each test name indicates expectations directly, no need for descriptive assertion failure message as well.

Please inline all calls to `checkForValidTopicFilter` and `checkForValidTopicFilter` and then remove them.

Also, no need for `topicFilter` variable in each test, so they each become one-liners, for example:
```
    @Test
    public void shouldNotValidateSingleLevelWildcardsCombinedWithTopicNameWithValidMultiLevelWildcard()
    {
        assertFalse(validator.isTopicFilterValid("+topic/name/#"));
    }
```

jfallows (changes requested):

jfallows (dismissed):

jfallows (approved):

PR #28, opened 2020-01-20
Updated all encoders to encode the correct amount of remainingLength bytes.

dknoma:
Also added properties fields to the suback and unsuback in protocol.idl as they can have properties even if they are not used at the moment.

### DIFF
--- a/src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
+++ b/src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
@@ -24,6 +24,7 @@ import static org.reaktivity.nukleus.mqtt.internal.MqttReasonCodes.MALFORMED_PAC
 import static org.reaktivity.nukleus.mqtt.internal.MqttReasonCodes.NORMAL_DISCONNECT;
 import static org.reaktivity.nukleus.mqtt.internal.MqttReasonCodes.PROTOCOL_ERROR;
 import static org.reaktivity.nukleus.mqtt.internal.MqttReasonCodes.SUCCESS;
+import static org.reaktivity.nukleus.mqtt.internal.MqttReasonCodes.TOPIC_FILTER_INVALID;
 import static org.reaktivity.nukleus.mqtt.internal.MqttReasonCodes.UNSUPPORTED_PROTOCOL_VERSION;
 import static org.reaktivity.nukleus.mqtt.internal.types.MqttRole.RECEIVER;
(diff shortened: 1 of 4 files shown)


src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1388,14 +1391,14 @@ private void doEncodePublish(
                 .correlationData(a -> a.bytes(dataEx.correlationInfo().bytes()))
                 .build();
 
-            final int limit = mqttPropertyRW.limit();
+            final int propertiesLimit = mqttPropertyRW.limit();
```
jfallows: This is representing `propertiesSize`, right?

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1430,7 +1434,7 @@ private void doEncodeSuback(
             for (int i = 0; i < ackCount; i++)
             {
                 final int ackIndex = 1 << i;
-                subscriptions[i] = (byte) ((successMask & ackIndex) > 0 ? 0x00 : 0x8F);
+                subscriptions[i] = (byte) ((successMask & ackIndex) > 0 ? SUCCESS : TOPIC_FILTER_INVALID);
```
jfallows: Better to use `!= 0` instead of `> 0` for bitmask checks.
Cast to `byte` can be avoided if constants are defined as type `byte` instead.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1439,9 +1443,10 @@ private void doEncodeSuback(
 
             final MqttSubackFW suback = mqttSubackRW.wrap(writeBuffer, DataFW.FIELD_OFFSET_PAYLOAD, writeBuffer.capacity())
                                             .typeAndFlags(0x90)
-                                            .remainingLength(reasonCodes.sizeof() + 1)
+                                            .remainingLength(3 + EMPTY_OCTETS.sizeof() + reasonCodes.sizeof())
```
jfallows, line 1446: `EMPTY_OCTETS.sizeof()` is always `0`, right?

jfallows (changes requested):

jfallows (approved):

PR #38, opened 2020-04-20
Verify network END, ABORT, and RESET

dknoma:
Added scripts to verify network reply is closed/aborted/reset via `onNetworkEnd`, `onNetworkAbort`, and `onNetworkReset`. Also tests that application streams are cleaned up and aborts properly in response to network reply.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -53,7 +53,7 @@
 
     <nukleus.plugin.version>0.49</nukleus.plugin.version>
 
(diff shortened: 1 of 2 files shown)


jfallows (changes requested):
Looks good, just need to rename IT methods when up-taking latest spec changes.

jfallows (approved):

PR #40, opened 2020-04-24
Implement CONNECT Keep Alive

dknoma:


src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1108,6 +1118,35 @@ private void onNetworkReset(
...
+        {
+            final long traceId = signal.traceId();
+            final long authorization = signal.authorization();
+
+            final long now = System.currentTimeMillis();
+            if (now >= keepAliveTimeoutAt)
+            {
+                onDecodeError(traceId, authorization, KEEP_ALIVE_TIMEOUT);
+                decoder = decodeIgnoreAll;
+            }
```
jfallows, line 1147: Need `else` clause to schedule keep-alive timeout signal at extended deadline?

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1242,6 +1283,7 @@ private void onDecodePublish(
 
                 final MqttDataExFW dataEx = builder.build();
                 stream.doApplicationData(traceId, authorization, reserved, payload, dataEx);
+                doSignalKeepAliveTimeoutIfNecessary();
```
jfallows: Just need to keep track of latest packet arrival time, for use in `onKeepAliveTimeoutSignal` to extend the deadline, same for other inbound network frames.

Initial signal will already be scheduled in `onDecodeConnect` and subsequent signals will be scheduled in `onKeepAliveTimeoutSignal` so don't need to trigger scheduling from here, right?

jfallows (changes requested):

jfallows (dismissed):

jfallows (approved):

PR #42, opened 2020-05-08
Update capabilities when PUBLISH comes after SUBSCRIBE

dknoma:
Allows streams to update their capabilities from `SUBSCRIBE_ONLY` to `PUBLISH_ONLY` or `PUBLISH_AND_SUBSCRIBE` when client sends `PUBLISH` some time after sending `SUBSCRIBE` for the same topic.

### DIFF
--- a/src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
+++ b/src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
@@ -640,13 +640,7 @@ public final class MqttServerFactory implements StreamFactory
                     }
                 }
 
-                assert publisher != null;
-
-                if (!hasPublishCapability(publisher.capabilities))
-                {
-                    publisher.capabilities |= PUBLISH_ONLY.value();
-                    publisher.doApplicationFlush(traceId, authorization, 0);
-                }
+                publisher.updatePublishCapabilityIfNecessary(traceId, authorization);
 
                 final OctetsFW payload = publish.payload();
                 final int payloadSize = payload.sizeof();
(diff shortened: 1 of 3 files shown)


src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -2607,6 +2595,29 @@ private void doCancelPublishExpirationIfNecessary()
...
+                int capabilities)
+            {
+                return (capabilities & PUBLISH_ONLY.value()) != 0;
+            }
+
+            private boolean hasSubscribeCapability(
+                int capabilities)
+            {
+                return (capabilities & SUBSCRIBE_ONLY.value()) != 0;
+            }
```
jfallows: Please move `hasPublishCapability` and `hasSubscribeCapability` to outer factory class, unless they are required to be here, but it doesn't seem so.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -2607,6 +2595,29 @@ private void doCancelPublishExpirationIfNecessary()
                     publishExpiresId = NO_CANCEL_ID;
                 }
             }
+
+            private void updatePublishCapabilityIfNecessary(
```
jfallows: Rename to `ensurePublishCapability`

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -640,7 +640,7 @@ private int decodePublish(
                     }
                 }
 
-                assert publisher != null;
+                publisher.updatePublishCapabilityIfNecessary(traceId, authorization);
```
jfallows: What happens if the subscribe stream is still active when the publisher times out?
Please check how the flow control behaves to make sure we are not sending `DATA` after a `FLUSH` that switched off `publish` capability without first receiving `WINDOW` after switching back on `publish` capability`.

jfallows (changes requested):

jfallows (approved):

PR #59, opened 2020-05-14
Feature/connect timeout

### LINKED ISSUE #53: Close network if CONNECT not received
See MQTT 5 specification:
- section [3.1.4 CONNECT Actions](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901073)

> If the Server does not receive a CONNECT packet within a reasonable amount of time after the Network Connection is established, the Server SHOULD close the Network Connection.

dknoma:
Fixes #53 

Server must end client connection if no `CONNECT` was sent within a configurable time frame from establishing the network connection.

Timeout is set via `MqttConfiguration`.

### DIFF
--- a/src/main/java/org/reaktivity/nukleus/mqtt/internal/MqttConfiguration.java
+++ b/src/main/java/org/reaktivity/nukleus/mqtt/internal/MqttConfiguration.java
@@ -45,17 +45,17 @@ public class MqttConfiguration extends Configuration
         super(MQTT_CONFIG, config);
     }
 
-    public long getPublishTimeout()
+    public long publishTimeout()
     {
         return PUBLISH_TIMEOUT.get(this);
     }
 
-    public long getConnectTimeout()
+    public long connectTimeout()
     {
         return CONNECT_TIMEOUT.get(this);
     }
 
-    public String getClientId()
+    public String clientId()
     {
         return CLIENT_ID.get(this);
     }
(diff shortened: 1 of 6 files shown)


src/main/java/org/reaktivity/nukleus/mqtt/internal/MqttConfiguration.java
```diff
@@ -47,6 +50,11 @@ public long getPublishTimeout()
         return PUBLISH_TIMEOUT.get(this);
     }
 
+    public long getConnectTimeout()
```
jfallows: Please remove the `get` prefix from all configuration methods.
So, `connectTimeout()`, `clientId()`, etc.

src/main/java/org/reaktivity/nukleus/mqtt/internal/MqttReasonCodes.java
```diff
@@ -56,6 +56,8 @@
     public static final byte PACKET_IDENTIFIER_IN_USE = -0x6F;
     public static final byte PACKET_IDENTIFIER_NOT_FOUND = -0x6E;
 
+    public static final byte MAXIMUM_CONNECT_TIME = -0x60;
+
```
jfallows: No longer required.

jfallows (changes requested):

jfallows (approved):

PR #62, opened 2020-05-21
Feature/topic alias

### LINKED ISSUE #43: Support topic alias for client PUBLISH
See MQTT 5 specification:
- section [3.2.2.3.8 Topic Alias Maximum](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901088)
- section [3.3.2.3.4 Topic Alias](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901113)
- section [3.3.4 PUBLISH Actions](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901120)

dknoma:
Fixes #43
scenarios
- use one alias
- use multiple distinct aliases
- use one alias, then overwrite it
- use multiple distinct aliases, then overwrite one
- publish with only `topic alias` to an existing `topic alias` mapping
- reject publish if `topic alias` > `topic alias maximum`

### DIFF
--- a/src/main/java/org/reaktivity/nukleus/mqtt/internal/MqttReasonCodes.java
+++ b/src/main/java/org/reaktivity/nukleus/mqtt/internal/MqttReasonCodes.java
@@ -33,30 +33,30 @@ public final class MqttReasonCodes
     public static final byte CONTINUE_AUTHENTICATION = 0x18;
     public static final byte REAUTHENTICATE = 0x19;
 
-    public static final byte UNSPECIFIED_ERROR = -0x80;
-    public static final byte MALFORMED_PACKET = -0x7F;
-    public static final byte PROTOCOL_ERROR = -0x7E;
-    public static final byte IMPLEMENTATION_SPECIFIC_ERROR = -0x7D;
-    public static final byte UNSUPPORTED_PROTOCOL_VERSION = -0x7C;
-    public static final byte CLIENT_IDENTIFIER_NOT_VALID = -0x7B;
-    public static final byte BAD_USER_NAME_OR_PASSWORD = -0x7A;
-    public static final byte NOT_AUTHORIZED = -0x79;
-    public static final byte SERVER_UNAVAILABLE = -0x78;
-    public static final byte SERVER_BUSY = -0x77;
-    public static final byte BANNED = -0x76;
-    public static final byte SERVER_SHUTTING_DOWN = -0x75;
-    public static final byte BAD_AUTHENTICATION_METHOD = -0x74;
+    public static final byte UNSPECIFIED_ERROR = (byte) 0x80;
+    public static final byte MALFORMED_PACKET = (byte) 0x81;
+    public static final byte PROTOCOL_ERROR = (byte) 0x82;
+    public static final byte IMPLEMENTATION_SPECIFIC_ERROR = (byte) 0x83;
+    public static final byte UNSUPPORTED_PROTOCOL_VERSION = (byte) 0x84;
+    public static final byte CLIENT_IDENTIFIER_NOT_VALID = (byte) 0x85;
+    public static final byte BAD_USER_NAME_OR_PASSWORD = (byte) 0x86;
+    public static final byte NOT_AUTHORIZED = (byte) 0x87;
+    public static final byte SERVER_UNAVAILABLE = (byte) 0x88;
+    public static final byte SERVER_BUSY = (byte) 0x89;
+    public static final byte BANNED = (byte) 0x8A;
+    public static final byte SERVER_SHUTTING_DOWN = (byte) 0x8B;
+    public static final byte BAD_AUTHENTICATION_METHOD = (byte) 0x8C;
 
-    public static final byte KEEP_ALIVE_TIMEOUT = -0x73;
-    public static final byte SESSION_TAKEN_OVER = -0x72;
+    public static final byte KEEP_ALIVE_TIMEOUT = (byte) 0x8D;
+    public static final byte SESSION_TAKEN_OVER = (byte) 0x8E;
 
-    public static final byte TOPIC_FILTER_INVALID = -0x71;
-    public static final byte TOPIC_NAME_INVALID = -0x70;
+    public static final byte TOPIC_FILTER_INVALID = (byte) 0x8F;
+    public static final byte TOPIC_NAME_INVALID = (byte) 0x90;
 
-    public static final byte PACKET_IDENTIFIER_IN_USE = -0x6F;
-    public static final byte PACKET_IDENTIFIER_NOT_FOUND = -0x6E;
-    public static final byte RECEIVE_MAXIMUM_EXCEEDED = -0x6D;
-    public static final byte TOPIC_ALIAS_INVALID = -0x6C;
+    public static final byte PACKET_IDENTIFIER_IN_USE = (byte) 0x91;
+    public static final byte PACKET_IDENTIFIER_NOT_FOUND = (byte) 0x92;
+    public static final byte RECEIVE_MAXIMUM_EXCEEDED = (byte) 0x93;
+    public static final byte TOPIC_ALIAS_INVALID = (byte) 0x94;
 
     private MqttReasonCodes()
     {
--- a/src/main/java/org/reaktivity/nukleus/mqtt/internal/MqttReasonCodes.java
+++ b/src/main/java/org/reaktivity/nukleus/mqtt/internal/MqttReasonCodes.java
@@ -55,6 +55,8 @@ public final class MqttReasonCodes
 
     public static final byte PACKET_IDENTIFIER_IN_USE = -0x6F;
     public static final byte PACKET_IDENTIFIER_NOT_FOUND = -0x6E;
+    public static final byte RECEIVE_MAXIMUM_EXCEEDED = -0x6D;
+    public static final byte TOPIC_ALIAS_INVALID = -0x6C;
 
     private MqttReasonCodes()
     {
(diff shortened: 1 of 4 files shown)


src/main/java/org/reaktivity/nukleus/mqtt/internal/MqttReasonCodes.java
```diff
@@ -55,6 +55,8 @@
 
     public static final byte PACKET_IDENTIFIER_IN_USE = -0x6F;
     public static final byte PACKET_IDENTIFIER_NOT_FOUND = -0x6E;
+    public static final byte RECEIVE_MAXIMUM_EXCEEDED = -0x6D;
+    public static final byte TOPIC_ALIAS_INVALID = -0x6C;
```
jfallows: Please change these literals to something like the following...
```
public static final byte PACKET_IDENTIFIER_IN_USE = (byte) 0x91;
```
...so that the values in the code line up more directly with the values in the MQTT v5 spec.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -619,58 +619,96 @@ private int decodePublish(
...
             }
-
-            if (reasonCode == 0)
+            else
             {
-                final int topicKey = topicKey(topicName);
-                MqttServer.MqttServerStream publisher = server.streams.get(topicKey);
-
-                if (publisher == null)
+                boolean hasTopicAlias = false;
```
jfallows: Track `alias` as `int` instead of creating a separate boolean flag variable.
Use initial value of `alias` that is illegal, such as `0`, so you can detect if `alias` was successfully set and therefore detect repetition.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -619,58 +619,96 @@ private int decodePublish(
...
+                        else
+                        {
+                            server.topicAliases.put(alias, topic);
+                        }
+                        break;
+                    case KIND_PAYLOAD_FORMAT:
+                        payloadFormat = MqttPayloadFormat.valueOf(mqttProperty.payloadFormat());
+                        break;
+                    case KIND_RESPONSE_TOPIC:
+                        responseTopic = mqttProperty.responseTopic().asString();
```
jfallows: Given that the underlying bytes are stable and remain on stack, can we avoid the `toString()` calls by wrapping the underlying bytes using a `String16FW responseTopicRO` on the factory, and passing that instead of `String` to `onDecodePublishSuccess`.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -682,6 +720,67 @@ private int decodePublish(
         return progress;
     }
 
+    private int onDecodePublishSuccess(
```
jfallows: This method is just part of the `decodePublish` method and can be inlined, right?
If the method is too long, then let's instead find a way to hoist the property decoding into a separate visitor to simplify the loop logic in `decodePublish`.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1204,6 +1307,39 @@ private void onDecodeConnect(
...
             }
 
+            final MqttPropertiesFW properties = packet.properties();
+
+            final OctetsFW propertiesValue = properties.value();
+            final DirectBuffer decodeBuffer = propertiesValue.buffer();
+            final int decodeOffset = propertiesValue.offset();
+            final int decodeLimit = propertiesValue.limit();
+
+            boolean topicAliasMaximumSet = false;
```
jfallows: Please track `topicAliasMaximum` with an initially illegal value to detect when it is set, and then remove this separate boolean tracking variable which is no longer needed.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1765,13 +1844,24 @@ private void doEncodeConnack(
             long authorization,
             int reasonCode)
         {
+            AtomicInteger propertiesSize = new AtomicInteger();
```
jfallows: Why do we need to instantiate here? Try to eliminate the allocation.

jfallows (changes requested):

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -605,41 +613,43 @@ private int decodePublish(
         final int limit)
     {
         final int length = limit - offset;
-
         int progress = offset;
 
-        decode:
```
jfallows, line 611: This `decode:` label is in place to support single return while allowing an easy escape from the outer if block.
Please restore it and use `break decode` in place of `return` mid method.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -2669,4 +2662,112 @@ private void ensurePublishCapability(
...
+        private void reset()
+        {
+            this.expiryInterval = DEFAULT_EXPIRY_INTERVAL;
+            this.contentType = null;
+            this.payloadFormat = DEFAULT_FORMAT;
+            this.responseTopic = null;
+            this.correlationData = null;
+        }
+
+        private void decodeProperties(MqttServer server, MqttPropertiesFW properties)
```
jfallows: Formatting of method arguments.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -2669,4 +2662,112 @@ private void ensurePublishCapability(
...
+                        if (mContentType.value() != null)
+                        {
+                            contentType = contentTypeRO.wrap(mContentType.buffer(), mContentType.offset(), mContentType.limit());
+                        }
+                        break;
+                    case KIND_TOPIC_ALIAS:
+                        if (alias != 0)
+                        {
+                            reasonCode = PROTOCOL_ERROR;
+                            break;
```
jfallows, line 2727: `break decode`

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -2669,4 +2662,112 @@ private void ensurePublishCapability(
...
+                        }
+
+                        alias = mqttProperty.topicAlias() & 0xFFFF;
+
+                        if (alias <= 0 || alias > server.topicAliasMaximum)
+                        {
+                            reasonCode = TOPIC_ALIAS_INVALID;
+                            break;
```
jfallows: `break decode`

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -2669,4 +2662,112 @@ private void ensurePublishCapability(
...
+                        }
+
+                        if (topic.isEmpty())
+                        {
+                            if (!server.topicAliases.containsKey(alias))
+                            {
+                                reasonCode = PROTOCOL_ERROR;
+                                break;
```
jfallows: `break decode`

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -2669,4 +2662,112 @@ private void ensurePublishCapability(
...
+                    case KIND_CORRELATION_DATA:
+                        correlationData = mqttProperty.correlationData().bytes();
+                        break;
+                    case KIND_USER_PROPERTY:
+                        final MqttUserPropertyFW userProperty = mqttProperty.userProperty();
+                        userPropertiesRW.item(c -> c.key(userProperty.key()).value(userProperty.value()));
+                        break;
+                    default:
+                        reasonCode = MALFORMED_PACKET;
+                        break;
```
jfallows: `break decode` ?
please check MQTT v5 spec to verify that unrecognized properties are to be rejected, not ignored

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -2669,4 +2662,112 @@ private void ensurePublishCapability(
...
+            final int decodeLimit = propertiesValue.limit();
+
+            if (topic == null)
+            {
+                reasonCode = PROTOCOL_ERROR;
+            }
+            else
+            {
+                int alias = 0;
+                for (int decodeProgress = decodeOffset; decodeProgress < decodeLimit; )
```
jfallows, line 2708: `decode:` label

jfallows (changes requested):

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
dknoma: For future reference, yes unrecognized properties should be rejected as `0x81 Malformed Packet`.

> When a Server detects a Malformed Packet or Protocol Error, and a Reason Code is given in the specification, it MUST close the Network Connection

jfallows (approved):

PR #63, opened 2020-05-30
Support server assigned client identifier

### LINKED ISSUE #48: Support server assigned client identifier
See MQTT 5 specification:
- section [3.2.2.3.7 Assigned Client Identifier](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901087)

dknoma:
fix #48 

Upon receiving an zero length client identifier on `CONNECT`, server must assign a unique client identifier to the client via `CONNACK`

### DIFF
--- a/src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
+++ b/src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
@@ -1204,8 +1204,17 @@ public final class MqttServerFactory implements StreamFactory
                 reasonCode = PROTOCOL_ERROR;
             }
 
+            final String clientIdentifier = packet.clientId().asString();
+
+            mqttPropertyRW.wrap(propertyBuffer, 0, propertyBuffer.capacity());
+
+            if (clientIdentifier != null && clientIdentifier.isEmpty())
+            {
+                mqttPropertyRW.assignedClientId(clientId);
+            }
+
             doCancelConnectTimeoutIfNecessary();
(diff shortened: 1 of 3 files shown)


src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1204,8 +1205,18 @@ private void onDecodeConnect(
                 reasonCode = PROTOCOL_ERROR;
             }
 
+            final String16FW clientIdentifier = packet.clientId();
+
+            mqttPropertyRW.wrap(propertyBuffer, 0, propertyBuffer.capacity());
```
jfallows: Use return value `mqttProperty` on stack rather than repeated references to field.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1527,7 +1538,7 @@ private void onDecodeError(
             }
             else
             {
-                doEncodeConnack(traceId, authorization, reasonCode);
+                doEncodeConnack(traceId, authorization, reasonCode, EMPTY_OCTETS.buffer(), 0, 0);
```
jfallows: Either `EMPTY_BUFFER` constant directly instead, or consider changing last 3 parameters of `doEncodeConnack` to a single `Flyweight` parameter instead.

jfallows (changes requested):

jfallows (approved):

Issue #44, opened 2020-05-10
Support topic alias for server PUBLISH


jfallows:
See MQTT 5 specification:
- section [3.1.2.11.5 Topic Alias Maximum](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901051)
- section [3.3.2.3.4 Topic Alias](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901113)
- section [3.3.4 PUBLISH Actions](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901120)

PR #73, opened 2020-08-26
CONNACK should match capabilities

### LINKED ISSUE #69: CONNACK should match capabilities
`CONNACK` error
 - Protocol version not supported `0x84`, if not `MQTT v5`
 - Bad authentication method `0x8c`, if specified

`CONNACK` properties
 - Session Expiry Interval `0`, see #55
 - Maximum QoS `0`, see #46 and #47
 - Retain Available `no`, see #54 
 - Wildcard Subscription Available `no`, see #50 
 - Subscription Identifier Available `yes`
 - Shared Subscription Available `no`, see #49
 - Topic Alias Maximum (verify), see #43 and #44

dknoma:
Fix #69
Requires https://github.com/reaktivity/nukleus-mqtt.spec/pull/44

### DIFF
--- a/src/main/java/org/reaktivity/nukleus/mqtt/internal/MqttConfiguration.java
+++ b/src/main/java/org/reaktivity/nukleus/mqtt/internal/MqttConfiguration.java
@@ -21,25 +21,14 @@ import org.reaktivity.nukleus.Configuration;
 
 public class MqttConfiguration extends Configuration
 {
-    public static final String PUBLISH_TIMEOUT_NAME = "nukleus.mqtt.publish.timeout";
-    public static final String CONNECT_TIMEOUT_NAME = "nukleus.mqtt.connect.timeout";
-    public static final String CLIENT_ID_NAME = "nukleus.mqtt.client.id";
-    public static final String RETAIN_AVAILABLE_NAME = "nukleus.mqtt.retain.available";
-    public static final String SESSION_EXPIRY_INTERVAL_NAME = "nukleus.mqtt.session.expiry.interval";
-    public static final String MAXIMUM_QOS_NAME = "nukleus.mqtt.maximum.qos";
-    public static final String TOPIC_ALIAS_MAXIMUM_AVAILABLE_NAME = "nukleus.mqtt.topic.alias.maximum.available";
-    public static final String WILDCARD_SUBSCRIPTION_AVAILABLE_NAME = "nukleus.mqtt.wildcard.subscription.available";
-    public static final String SUBSCRIPTION_IDENTIFIERS_AVAILABLE_NAME = "nukleus.mqtt.subscription.identifiers.available";
-    public static final String SHARED_SUBSCRIPTION_AVAILABLE_NAME = "nukleus.mqtt.shared.subscription.available";
-
     private static final ConfigurationDef MQTT_CONFIG;
     public static final LongPropertyDef CONNECT_TIMEOUT;
     public static final LongPropertyDef PUBLISH_TIMEOUT;
     public static final PropertyDef<String> CLIENT_ID;
-    public static final BooleanPropertyDef RETAIN_AVAILABLE;
     public static final IntPropertyDef SESSION_EXPIRY_INTERVAL;
-    public static final BytePropertyDef MAXIMUM_QOS;    public static final BooleanPropertyDef TOPIC_ALIAS_MAXIMUM_AVAILABLE;
-
+    public static final BytePropertyDef MAXIMUM_QOS;
+    public static final BooleanPropertyDef RETAIN_AVAILABLE;
+    public static final ShortPropertyDef TOPIC_ALIAS_MAXIMUM;
     public static final BooleanPropertyDef WILDCARD_SUBSCRIPTION_AVAILABLE;
     public static final BooleanPropertyDef SUBSCRIPTION_IDENTIFIERS_AVAILABLE;
     public static final BooleanPropertyDef SHARED_SUBSCRIPTION_AVAILABLE;
@@ -50,10 +39,10 @@ public class MqttConfiguration extends Configuration
         PUBLISH_TIMEOUT = config.property("publish.timeout", TimeUnit.SECONDS.toSeconds(30));
         CONNECT_TIMEOUT = config.property("connect.timeout", TimeUnit.SECONDS.toSeconds(3));
         CLIENT_ID = config.property("client.id", "client");
-        RETAIN_AVAILABLE = config.property("retain.available", true);
         SESSION_EXPIRY_INTERVAL = config.property("session.expiry.interval", 0);
         MAXIMUM_QOS = config.property("maximum.qos", (byte) 0);
-        TOPIC_ALIAS_MAXIMUM_AVAILABLE = config.property("topic.alias.maximum.available", false);
+        RETAIN_AVAILABLE = config.property("retain.available", true);
+        TOPIC_ALIAS_MAXIMUM = config.property("topic.alias.maximum", (short) 0);
         WILDCARD_SUBSCRIPTION_AVAILABLE = config.property("wildcard.subscription.available", false);
         SUBSCRIPTION_IDENTIFIERS_AVAILABLE = config.property("subscription.identifiers.available", true);
         SHARED_SUBSCRIPTION_AVAILABLE = config.property("shared.subscription.available", false);
@@ -96,9 +85,9 @@ public class MqttConfiguration extends Configuration
         return MAXIMUM_QOS.get(this);
     }
 
-    public boolean topicAliasMaximumAvailable()
+    public short topicAliasMaximum()
     {
-        return TOPIC_ALIAS_MAXIMUM_AVAILABLE.get(this);
+        return TOPIC_ALIAS_MAXIMUM.get(this);
     }
 
     public boolean wildcardSubscriptionAvailable()
--- a/src/main/java/org/reaktivity/nukleus/mqtt/internal/MqttConfiguration.java
+++ b/src/main/java/org/reaktivity/nukleus/mqtt/internal/MqttConfiguration.java
@@ -27,10 +27,10 @@ public class MqttConfiguration extends Configuration
     public static final String RETAIN_AVAILABLE_NAME = "nukleus.mqtt.retain.available";
     public static final String SESSION_EXPIRY_INTERVAL_NAME = "nukleus.mqtt.session.expiry.interval";
     public static final String MAXIMUM_QOS_NAME = "nukleus.mqtt.maximum.qos";
+    public static final String TOPIC_ALIAS_MAXIMUM_AVAILABLE_NAME = "nukleus.mqtt.topic.alias.maximum.available";
     public static final String WILDCARD_SUBSCRIPTION_AVAILABLE_NAME = "nukleus.mqtt.wildcard.subscription.available";
     public static final String SUBSCRIPTION_IDENTIFIERS_AVAILABLE_NAME = "nukleus.mqtt.subscription.identifiers.available";
     public static final String SHARED_SUBSCRIPTION_AVAILABLE_NAME = "nukleus.mqtt.shared.subscription.available";
-    public static final String TOPIC_ALIAS_MAXIMUM_AVAILABLE_NAME = "nukleus.mqtt.topic.alias.maximum.available";
 
     private static final ConfigurationDef MQTT_CONFIG;
     public static final LongPropertyDef CONNECT_TIMEOUT;
@@ -38,11 +38,11 @@ public class MqttConfiguration extends Configuration
     public static final PropertyDef<String> CLIENT_ID;
     public static final BooleanPropertyDef RETAIN_AVAILABLE;
     public static final IntPropertyDef SESSION_EXPIRY_INTERVAL;
-    public static final BytePropertyDef MAXIMUM_QOS;
+    public static final BytePropertyDef MAXIMUM_QOS;    public static final BooleanPropertyDef TOPIC_ALIAS_MAXIMUM_AVAILABLE;
+
     public static final BooleanPropertyDef WILDCARD_SUBSCRIPTION_AVAILABLE;
     public static final BooleanPropertyDef SUBSCRIPTION_IDENTIFIERS_AVAILABLE;
     public static final BooleanPropertyDef SHARED_SUBSCRIPTION_AVAILABLE;
-    public static final BooleanPropertyDef TOPIC_ALIAS_MAXIMUM_AVIALABLE;
 
     static
     {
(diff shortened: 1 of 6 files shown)


src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1772,26 +1787,56 @@ private void doEncodeConnack(
...
+
+            mqttPropertyRW.wrap(propertyBuffer, propertiesSize, propertyBuffer.capacity())
+                          .subscriptionIdsAvailable(subscriptionIdentifierAvailable)
+                          .build();
+            propertiesSize = mqttPropertyRW.limit();
+
+            mqttPropertyRW.wrap(propertyBuffer, propertiesSize, propertyBuffer.capacity())
+                          .sharedSubscriptionAvailable(sharedSubscriptionAvailable)
+                          .build();
+            propertiesSize = mqttPropertyRW.limit();
```
jfallows: In each case, test for the default setting, and omit the property if it would convey the right value implicitly, not just for topic alias maximum.

src/main/java/org/reaktivity/nukleus/mqtt/internal/MqttConfiguration.java
```diff
@@ -25,12 +25,24 @@
     public static final String CONNECT_TIMEOUT_NAME = "nukleus.mqtt.connect.timeout";
     public static final String CLIENT_ID_NAME = "nukleus.mqtt.client.id";
     public static final String RETAIN_AVAILABLE_NAME = "nukleus.mqtt.retain.available";
+    public static final String SESSION_EXPIRY_INTERVAL_NAME = "nukleus.mqtt.session.expiry.interval";
+    public static final String MAXIMUM_QOS_NAME = "nukleus.mqtt.maximum.qos";
+    public static final String TOPIC_ALIAS_MAXIMUM_AVAILABLE_NAME = "nukleus.mqtt.topic.alias.maximum.available";
+    public static final String WILDCARD_SUBSCRIPTION_AVAILABLE_NAME = "nukleus.mqtt.wildcard.subscription.available";
+    public static final String SUBSCRIPTION_IDENTIFIERS_AVAILABLE_NAME = "nukleus.mqtt.subscription.identifiers.available";
+    public static final String SHARED_SUBSCRIPTION_AVAILABLE_NAME = "nukleus.mqtt.shared.subscription.available";
```
jfallows: Consider moving these `_NAME` constants to `ConfigurationTest` as they are only used by test code.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -281,6 +289,13 @@ public MqttServerFactory(
         this.clientId = config.clientId();
         this.publishTimeoutMillis = SECONDS.toMillis(config.publishTimeout());
         this.connectTimeoutMillis = SECONDS.toMillis(config.connectTimeout());
+        this.sessionExpiryInterval = config.sessionExpiryInterval();
+        this.maximumQos = config.maximumQos();
+        this.retainAvailable = config.retainAvailable() ? (byte) 1 : 0;
+        this.wildcardSubscriptionAvailable = config.wildcardSubscriptionAvailable() ? (byte) 1 : 0;
+        this.subscriptionIdentifierAvailable = config.subscriptionIdentifierAvailable() ? (byte) 1 : 0;
+        this.sharedSubscriptionAvailable = config.sharedSubscriptionAvailable() ? (byte) 1 : 0;
+        this.topicAliasMaximumAvailable = config.topicAliasMaximumAvailable();
```
jfallows: How about changing the configuration to indicate the actual topic alias maximum, rather than a boolean?

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -281,6 +289,13 @@ public MqttServerFactory(
         this.clientId = config.clientId();
         this.publishTimeoutMillis = SECONDS.toMillis(config.publishTimeout());
         this.connectTimeoutMillis = SECONDS.toMillis(config.connectTimeout());
+        this.sessionExpiryInterval = config.sessionExpiryInterval();
+        this.maximumQos = config.maximumQos();
+        this.retainAvailable = config.retainAvailable() ? (byte) 1 : 0;
+        this.wildcardSubscriptionAvailable = config.wildcardSubscriptionAvailable() ? (byte) 1 : 0;
+        this.subscriptionIdentifierAvailable = config.subscriptionIdentifierAvailable() ? (byte) 1 : 0;
+        this.sharedSubscriptionAvailable = config.sharedSubscriptionAvailable() ? (byte) 1 : 0;
```
jfallows: Simplify `retainAvailable`, `wildcardSubscriptionAvailable`, `subscriptionIdentifierAvailable` and `sharedSubscriptionAvailable` to booleans called `retainedMessages`, `wildcardSubscriptions`, `subscriptionIdentifiers` and `sharedSubscriptions`?

jfallows (changes requested):

src/main/java/org/reaktivity/nukleus/mqtt/internal/MqttConfiguration.java
```diff
@@ -21,24 +21,33 @@
...
+    public static final BooleanPropertyDef SHARED_SUBSCRIPTION_AVAILABLE;
+    public static final BooleanPropertyDef NO_LOCAL;
 
     static
     {
         final ConfigurationDef config = new ConfigurationDef("nukleus.mqtt");
         PUBLISH_TIMEOUT = config.property("publish.timeout", TimeUnit.SECONDS.toSeconds(30));
         CONNECT_TIMEOUT = config.property("connect.timeout", TimeUnit.SECONDS.toSeconds(3));
         CLIENT_ID = config.property("client.id", "client");
+        SESSION_EXPIRY_INTERVAL = config.property("session.expiry.interval", -1);
```
jfallows: Shouldn't this default to `0` meaning `expire-on-network-close` instead of `-1`?

src/main/java/org/reaktivity/nukleus/mqtt/internal/MqttConfiguration.java
```diff
@@ -21,24 +21,33 @@
...
+        MAXIMUM_QOS = config.property("maximum.qos", (byte) -1);
```
jfallows: Shouldn't this default to `0` since we don't support a higher maximum `qos` yet?

src/main/java/org/reaktivity/nukleus/mqtt/internal/MqttConfiguration.java
```diff
@@ -21,24 +21,33 @@
...
         RETAIN_AVAILABLE = config.property("retain.available", true);
+        TOPIC_ALIAS_MAXIMUM = config.property("topic.alias.maximum", (short) -1);
+        WILDCARD_SUBSCRIPTION_AVAILABLE = config.property("wildcard.subscription.available", false);
+        SUBSCRIPTION_IDENTIFIERS_AVAILABLE = config.property("subscription.identifiers.available", true);
+        SHARED_SUBSCRIPTION_AVAILABLE = config.property("shared.subscription.available", false);
+        NO_LOCAL = config.property("no.local", true);
```
jfallows: Shouldn't this default to `false` as we don't support `no-local` yet?

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -599,19 +637,34 @@ private int decodeConnect(
...
 
             final MqttConnectFW mqttConnect = mqttConnectRO.tryWrap(buffer, offset, limit);
+            int flags;
             if (mqttConnect == null)
             {
                 reasonCode = PROTOCOL_ERROR;
             }
             else if ((mqttConnect.typeAndFlags() & 0b1111_1111) != CONNECT_FIXED_HEADER ||
-                     (mqttConnect.flags() & 0b0000_0001) != 0b0000_0000)
+                     ((flags = mqttConnect.flags()) & 0b0000_0001) != 0b0000_0000)
```
jfallows: Avoid putting side effects into conditions, it makes the code harder to reason about to verify correctness in all cases.

Instead we normally have a `decode:` label starting the block, with a `break decode;` where needed to escape the block. This makes it easier to have code between blocks, without needing nested `if-else` blocks.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -629,6 +682,12 @@ else if (!"MQTT".equals(mqttConnect.protocolName().asString()) || mqttConnect.pr
         return progress;
     }
 
+    private boolean checkWillFlag(
```
jfallows: Suggest renaming to `isSetWillFlag` and making this `static` at the end of the factory class.
Suggest adding other similar methods to manage flags and make the conditional logic above more readable.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -2749,6 +2878,160 @@ private void ensurePublishCapability(
...
+            MqttConnectFW connect)
+        {
+            final int flags = connect.flags();
+            final OctetsFW payload = connect.payload();
+
+            final DirectBuffer buffer = payload.buffer();
+            final int offset = payload.offset();
+            final int limit = payload.limit();
+
+            int payloadOffset = offset;
```
jfallows: Rename `payloadOffset` to `progress`;

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -2749,6 +2878,160 @@ private void ensurePublishCapability(
...
+            this.willPropertiesRO.reset();
+            this.willTopic = null;
+            this.willPayload = null;
+            this.username = null;
+            this.password = null;
+        }
+
+        private void decodePayload(
+            MqttConnectFW connect)
+        {
```
jfallows, line 2942: Suggest combining these 2 methods into `decode(MqttConnectFW)`, and adding a return type of `MqttConnectPayload`.

Then the caller will do something like:
```
MqttConnectPayload payload = mqttConnectPayloadRO.decode(connect);
payload.reasonCode ...
```

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -2749,6 +2878,160 @@ private void ensurePublishCapability(
...
+            this.payloadFormat = DEFAULT_FORMAT;
+            this.responseTopic = null;
+            this.correlationData = null;
+        }
+
+        private void decodeProperties(
+            DirectBuffer payloadBuffer,
+            int payloadOffset,
+            int payloadLimit)
+        {
```
jfallows, line 3035: Suggest combining these 2 methods into `decode(MqttConnectFW)`, and adding a return type of `MqttConnectPayload`.

Then the caller will do something like:
```
MqttWillProperties willProperties = mqttWillPropertiesRO.decode(buffer, offset, limit);
willProperties.topic ...
```

This would be a useful pattern to apply to other payload decoders as well.

jfallows (changes requested):

src/main/java/org/reaktivity/nukleus/mqtt/internal/MqttConfiguration.java
```diff
@@ -21,24 +21,33 @@
...
     static
     {
         final ConfigurationDef config = new ConfigurationDef("nukleus.mqtt");
         PUBLISH_TIMEOUT = config.property("publish.timeout", TimeUnit.SECONDS.toSeconds(30));
         CONNECT_TIMEOUT = config.property("connect.timeout", TimeUnit.SECONDS.toSeconds(3));
         CLIENT_ID = config.property("client.id", "client");
+        SESSION_EXPIRY_INTERVAL = config.property("session.expiry.interval", 0);
+        MAXIMUM_QOS = config.property("maximum.qos", (byte) 0);
         RETAIN_AVAILABLE = config.property("retain.available", true);
+        TOPIC_ALIAS_MAXIMUM = config.property("topic.alias.maximum", (short) -1);
```
jfallows: If we don't support `topic alias` then this should default to `0`, if we do then it should default to the size we support by default rather than `-1`.

Then when encoding, the `topic alias maximum` should be omitted if `0`, as that is the spec default.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1267,15 +1324,22 @@ private void onDecodeConnect(
                         reasonCode = PROTOCOL_ERROR;
                         break decode;
                     }
-                    topicAliasMaximum = (short) (mqttProperty.topicAliasMaximum() & 0xFFFF);
+                    final short topicAliasMaximum = (short) (mqttProperty.topicAliasMaximum() & 0xFFFF);
+                    this.topicAliasMaximum = topicAliasMaximumDefault >= 0 && topicAliasMaximum > topicAliasMaximumDefault ?
+                                                 topicAliasMaximumDefault :
+                                                 topicAliasMaximum;
```
jfallows: After changing `topicAliasMaximumLimit` to be `>= 0` then this becomes
```
this.topicAliasMaximum = Math.min(topicAliasMaximum, topicAliasMaximumLimit);
```

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -298,6 +330,14 @@ public MqttServerFactory(
         this.clientId = config.clientId();
         this.publishTimeoutMillis = SECONDS.toMillis(config.publishTimeout());
         this.connectTimeoutMillis = SECONDS.toMillis(config.connectTimeout());
+        this.sessionExpiryInterval = config.sessionExpiryInterval();
+        this.maximumQos = config.maximumQos();
+        this.retainedMessages = config.retainAvailable() ? (byte) 1 : 0;
+        this.wildcardSubscriptions = config.wildcardSubscriptionAvailable() ? (byte) 1 : 0;
+        this.subscriptionIdentifiers = config.subscriptionIdentifierAvailable() ? (byte) 1 : 0;
+        this.sharedSubscriptions = config.sharedSubscriptionAvailable() ? (byte) 1 : 0;
+        this.topicAliasMaximumDefault = config.topicAliasMaximum();
```
jfallows: Suggest renaming to `topicAliasMaximumLimit` and also enforcing `>= 0` giving:
```
this.topicAliasMaximumLimit = Math.max(config.topicAliasMaximum(), 0);
```

jfallows (changes requested):

jfallows (dismissed):

jfallows (approved):

Issue #45, opened 2020-05-10
Cleanup failed SUBSCRIBE application stream


jfallows:
If `SUBSCRIBE` fails, then subsequent `PUBLISH` still tries to use defunct application stream because it has not been cleaned up.

jfallows:
No longer reproduces and spec ITs now added to cover scenario.

PR #74, opened 2020-09-17
Support WILL message

### LINKED ISSUE #56: Support WILL message
See MQTT 5 specification:
- section [3.1.2.5 Will Flag](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901040)
- section [3.1.2.6 Will QoS](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901041)
- section [3.1.2.7 Will Retain](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901042)
- section [3.1.3.2 Will Properties](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901060)
- section [3.1.3.3 Will Topic](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901069)
- section [3.1.3.4 Will Payload](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901070)

Consider using an internal `$SYS/sessions/<client-id>` topic to manage will messages.

dknoma:
Fix #56 
Requires https://github.com/reaktivity/nukleus-mqtt.spec/pull/45

### DIFF
--- a/src/main/java/org/reaktivity/nukleus/mqtt/internal/MqttConfiguration.java
+++ b/src/main/java/org/reaktivity/nukleus/mqtt/internal/MqttConfiguration.java
@@ -33,7 +33,7 @@ public class MqttConfiguration extends Configuration
     public static final BooleanPropertyDef SUBSCRIPTION_IDENTIFIERS_AVAILABLE;
     public static final BooleanPropertyDef SHARED_SUBSCRIPTION_AVAILABLE;
     public static final BooleanPropertyDef NO_LOCAL;
-    public static final BooleanPropertyDef SESSIONS_AVAILABLE;
+    public static final IntPropertyDef SESSION_EXPIRY_GRACE_PERIOD;
 
     static
     {
@@ -49,7 +49,7 @@ public class MqttConfiguration extends Configuration
         SUBSCRIPTION_IDENTIFIERS_AVAILABLE = config.property("subscription.identifiers.available", true);
         SHARED_SUBSCRIPTION_AVAILABLE = config.property("shared.subscription.available", false);
         NO_LOCAL = config.property("no.local", false);
-        SESSIONS_AVAILABLE = config.property("sessions.available", true);
+        SESSION_EXPIRY_GRACE_PERIOD = config.property("session.expiry.grace.period", 0);
         MQTT_CONFIG = config;
     }
 
@@ -114,8 +114,8 @@ public class MqttConfiguration extends Configuration
         return NO_LOCAL.get(this);
     }
 
-    public boolean sessionsAvailable()
+    public int sessionExpiryUpdateInterval()
     {
-        return SESSIONS_AVAILABLE.get(this);
+        return SESSION_EXPIRY_GRACE_PERIOD.get(this);
     }
 }
--- a/src/main/java/org/reaktivity/nukleus/mqtt/internal/MqttConfiguration.java
+++ b/src/main/java/org/reaktivity/nukleus/mqtt/internal/MqttConfiguration.java
@@ -114,7 +114,7 @@ public class MqttConfiguration extends Configuration
         return NO_LOCAL.get(this);
     }
 
-    public int sessionExpiryUpdateInterval()
+    public int sessionExpiryGracePeriod()
     {
         return SESSION_EXPIRY_GRACE_PERIOD.get(this);
     }
--- a/src/main/java/org/reaktivity/nukleus/mqtt/internal/MqttConfiguration.java
+++ b/src/main/java/org/reaktivity/nukleus/mqtt/internal/MqttConfiguration.java
@@ -49,7 +49,7 @@ public class MqttConfiguration extends Configuration
         SUBSCRIPTION_IDENTIFIERS_AVAILABLE = config.property("subscription.identifiers.available", true);
         SHARED_SUBSCRIPTION_AVAILABLE = config.property("shared.subscription.available", false);
         NO_LOCAL = config.property("no.local", false);
-        SESSION_EXPIRY_GRACE_PERIOD = config.property("session.expiry.grace.period", 0);
+        SESSION_EXPIRY_GRACE_PERIOD = config.property("session.expiry.grace.period", 30);
         MQTT_CONFIG = config;
     }
 
--- a/src/main/java/org/reaktivity/nukleus/mqtt/internal/MqttConfiguration.java
+++ b/src/main/java/org/reaktivity/nukleus/mqtt/internal/MqttConfiguration.java
@@ -33,6 +33,7 @@ public class MqttConfiguration extends Configuration
     public static final BooleanPropertyDef SUBSCRIPTION_IDENTIFIERS_AVAILABLE;
     public static final BooleanPropertyDef SHARED_SUBSCRIPTION_AVAILABLE;
     public static final BooleanPropertyDef NO_LOCAL;
+    public static final BooleanPropertyDef SESSIONS_AVAILABLE;
 
     static
     {
@@ -48,6 +49,7 @@ public class MqttConfiguration extends Configuration
         SUBSCRIPTION_IDENTIFIERS_AVAILABLE = config.property("subscription.identifiers.available", true);
         SHARED_SUBSCRIPTION_AVAILABLE = config.property("shared.subscription.available", false);
         NO_LOCAL = config.property("no.local", false);
+        SESSIONS_AVAILABLE = config.property("sessions.available", true);
         MQTT_CONFIG = config;
     }
 
@@ -110,5 +112,10 @@ public class MqttConfiguration extends Configuration
     public boolean noLocal()
     {
         return NO_LOCAL.get(this);
+    }
+
+    public boolean sessionsAvailable()
+    {
+        return SESSIONS_AVAILABLE.get(this);
     }
 }
--- a/src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
+++ b/src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
@@ -58,7 +58,6 @@ import static org.reaktivity.nukleus.mqtt.internal.types.codec.MqttPropertyFW.KI
 import static org.reaktivity.nukleus.mqtt.internal.types.codec.MqttPropertyFW.KIND_TOPIC_ALIAS;
 import static org.reaktivity.nukleus.mqtt.internal.types.codec.MqttPropertyFW.KIND_TOPIC_ALIAS_MAXIMUM;
 import static org.reaktivity.nukleus.mqtt.internal.types.codec.MqttPropertyFW.KIND_USER_PROPERTY;
-import static org.reaktivity.nukleus.mqtt.internal.types.codec.MqttPropertyFW.KIND_WILL_DELAY_INTERVAL;
 import static org.reaktivity.nukleus.mqtt.internal.types.stream.DataFW.FIELD_OFFSET_PAYLOAD;
 import static org.reaktivity.nukleus.mqtt.internal.types.stream.MqttDataExFW.Builder.DEFAULT_EXPIRY_INTERVAL;
 import static org.reaktivity.nukleus.mqtt.internal.types.stream.MqttDataExFW.Builder.DEFAULT_FORMAT;
@@ -105,14 +104,17 @@ import org.reaktivity.nukleus.mqtt.internal.types.codec.MqttPingRespFW;
 import org.reaktivity.nukleus.mqtt.internal.types.codec.MqttPropertiesFW;
 import org.reaktivity.nukleus.mqtt.internal.types.codec.MqttPropertyFW;
 import org.reaktivity.nukleus.mqtt.internal.types.codec.MqttPublishFW;
+import org.reaktivity.nukleus.mqtt.internal.types.codec.MqttSessionFW;
 import org.reaktivity.nukleus.mqtt.internal.types.codec.MqttSubackFW;
 import org.reaktivity.nukleus.mqtt.internal.types.codec.MqttSubscribeFW;
 import org.reaktivity.nukleus.mqtt.internal.types.codec.MqttSubscribePayloadFW;
+import org.reaktivity.nukleus.mqtt.internal.types.codec.MqttSubscriptionsFW;
 import org.reaktivity.nukleus.mqtt.internal.types.codec.MqttUnsubackFW;
 import org.reaktivity.nukleus.mqtt.internal.types.codec.MqttUnsubackPayloadFW;
 import org.reaktivity.nukleus.mqtt.internal.types.codec.MqttUnsubscribeFW;
 import org.reaktivity.nukleus.mqtt.internal.types.codec.MqttUnsubscribePayloadFW;
 import org.reaktivity.nukleus.mqtt.internal.types.codec.MqttUserPropertyFW;
+import org.reaktivity.nukleus.mqtt.internal.types.codec.MqttWillFW;
 import org.reaktivity.nukleus.mqtt.internal.types.control.MqttRouteExFW;
 import org.reaktivity.nukleus.mqtt.internal.types.control.RouteFW;
 import org.reaktivity.nukleus.mqtt.internal.types.stream.AbortFW;
@@ -132,6 +134,8 @@ import org.reaktivity.nukleus.stream.StreamFactory;
 public final class MqttServerFactory implements StreamFactory
 {
     private static final OctetsFW EMPTY_OCTETS = new OctetsFW().wrap(new UnsafeBuffer(new byte[0]), 0, 0);
+
+    private static final String SESSION_TOPIC_FORMAT = "$SYS/sessions/%s";
 
     private static final int CONNECT_FIXED_HEADER = 0b0001_0000;
     private static final int SUBSCRIBE_FIXED_HEADER = 0b1000_0010;
@@ -166,6 +170,8 @@ public final class MqttServerFactory implements StreamFactory
     private static final int CONNECT_TIMEOUT_SIGNAL = 3;
 
     private static final int PUBLISH_FRAMING = 255;
+
+    private static final String16FW NULL_STRING = new String16FW((String) null);
 
     private final RouteFW routeRO = new RouteFW();
     private final MqttRouteExFW mqttRouteExRO = new MqttRouteExFW();
@@ -219,7 +225,6 @@ public final class MqttServerFactory implements StreamFactory
 
     private final MqttPublishHeader mqttPublishHeaderRO = new MqttPublishHeader();
     private final MqttConnectPayload mqttConnectPayloadRO = new MqttConnectPayload();
-    private final MqttWillProperties willPropertiesRO = new MqttWillProperties();
 
     private final MqttConnackFW.Builder mqttConnackRW = new MqttConnackFW.Builder();
     private final MqttPublishFW.Builder mqttPublishRW = new MqttPublishFW.Builder();
@@ -231,6 +236,10 @@ public final class MqttServerFactory implements StreamFactory
     private final Array32FW.Builder<MqttUserPropertyFW.Builder, MqttUserPropertyFW> userPropertiesRW =
         new Array32FW.Builder<>(new MqttUserPropertyFW.Builder(), new MqttUserPropertyFW());
 
+    private final MqttSubscriptionsFW.Builder subscriptionsRW = new MqttSubscriptionsFW.Builder();
+    private final MqttWillFW.Builder willRW = new MqttWillFW.Builder();
+    private final MqttSessionFW.Builder sessionRW = new MqttSessionFW.Builder();
+
     private final Signaler signaler;
 
     private final RouteManager router;
@@ -240,6 +249,9 @@ public final class MqttServerFactory implements StreamFactory
     private final MutableDirectBuffer payloadBuffer;
     private final MutableDirectBuffer propertyBuffer;
     private final MutableDirectBuffer userPropertiesBuffer;
+    private final MutableDirectBuffer sessionBuffer;
+    private final MutableDirectBuffer subscriptionsBuffer;
+    private final MutableDirectBuffer willBuffer;
     private final LongUnaryOperator supplyInitialId;
     private final LongUnaryOperator supplyReplyId;
     private final LongSupplier supplyTraceId;
@@ -250,7 +262,7 @@ public final class MqttServerFactory implements StreamFactory
     private final long connectTimeoutMillis;
     private final int encodeBudgetMax;
 
-    private final int sessionExpiryInterval;
+    private final int sessionExpiryIntervalLimit;
     private final byte maximumQos;
     private final byte retainedMessages;
     private final short topicAliasMaximumLimit;
@@ -317,6 +329,9 @@ public final class MqttServerFactory implements StreamFactory
         this.propertyBuffer = new UnsafeBuffer(new byte[writeBuffer.capacity()]);
         this.userPropertiesBuffer = new UnsafeBuffer(new byte[writeBuffer.capacity()]);
         this.payloadBuffer = new UnsafeBuffer(new byte[writeBuffer.capacity()]);
+        this.sessionBuffer = new UnsafeBuffer(new byte[writeBuffer.capacity()]);
+        this.subscriptionsBuffer = new UnsafeBuffer(new byte[writeBuffer.capacity()]);
+        this.willBuffer = new UnsafeBuffer(new byte[writeBuffer.capacity()]);
         this.bufferPool = bufferPool;
         this.creditor = creditor;
         this.supplyDebitor = supplyDebitor;
@@ -330,7 +345,7 @@ public final class MqttServerFactory implements StreamFactory
         this.clientId = config.clientId();
         this.publishTimeoutMillis = SECONDS.toMillis(config.publishTimeout());
         this.connectTimeoutMillis = SECONDS.toMillis(config.connectTimeout());
-        this.sessionExpiryInterval = config.sessionExpiryInterval();
+        this.sessionExpiryIntervalLimit = config.sessionExpiryInterval();
         this.maximumQos = config.maximumQos();
         this.retainedMessages = config.retainAvailable() ? (byte) 1 : 0;
         this.wildcardSubscriptions = config.wildcardSubscriptionAvailable() ? (byte) 1 : 0;
@@ -1034,6 +1049,8 @@ public final class MqttServerFactory implements StreamFactory
         private boolean connected;
 
         private short topicAliasMaximum = -1;
+        private int sessionExpiryInterval = -1;
+        private long sessionExpiresAt = 0;
 
         private int state;
 
@@ -1328,6 +1345,16 @@ public final class MqttServerFactory implements StreamFactory
                     this.topicAliasMaximum = (short) Math.min(topicAliasMaximum, topicAliasMaximumLimit);
                     break;
                 case KIND_SESSION_EXPIRY:
+                    if (sessionExpiryInterval != -1)
+                    {
+                        sessionExpiryInterval = 0;
+                        reasonCode = PROTOCOL_ERROR;
+                        break decode;
+                    }
+                    final int sessionExpiryInterval = mqttProperty.expiryInterval();
+                    this.sessionExpiryInterval = Math.min(sessionExpiryInterval, sessionExpiryIntervalLimit);
+                    this.sessionExpiresAt = System.currentTimeMillis() + sessionExpiryInterval * 1000;
+                    break;
                 case KIND_RECEIVE_MAXIMUM:
                 case KIND_MAXIMUM_PACKET_SIZE:
                 case KIND_REQUEST_RESPONSE_INFORMATION:
@@ -1348,6 +1375,20 @@ public final class MqttServerFactory implements StreamFactory
 
                 decodeProgress = mqttProperty.limit();
             }
+
+            sessionRW.wrap(sessionBuffer, 0, sessionBuffer.capacity())
+                     .expiresAt(sessionExpiresAt);
+
+            subscriptionsRW.wrap(subscriptionsBuffer, 0, subscriptionsBuffer.capacity())
+                           .subscriptions(ss -> streams.forEach((tk, s) ->
+                           {
+                               if (hasSubscribeCapability(s.capabilities))
+                               {
+                                   ss.item(i -> i.topic(s.topicFilter)
+                                                 .identifier(s.subscription.id)
+                                                 .flags(s.subscription.flags));
+                               }
+                           }));
 
             final String16FW clientIdentifier = packet.clientId();
 
@@ -1529,6 +1570,17 @@ public final class MqttServerFactory implements StreamFactory
                         unrouteableMask |= 1 << subscription.ackCount;
                     }
                 }
+
+                subscriptionsRW.wrap(subscriptionsBuffer, 0, subscriptionsBuffer.capacity())
+                               .subscriptions(ss -> streams.forEach((tk, s) ->
+                               {
+                                   if (hasSubscribeCapability(s.capabilities))
+                                   {
+                                       ss.item(i -> i.topic(s.topicFilter)
+                                                     .identifier(s.subscription.id)
+                                                     .flags(s.subscription.flags));
+                                   }
+                               }));
 
                 subscription.ackMask |= unrouteableMask;
                 doSignalKeepAliveTimeoutIfNecessary();
@@ -1872,10 +1924,10 @@ public final class MqttServerFactory implements StreamFactory
         {
             int propertiesSize = 0;
 
-            if (sessionExpiryInterval > 0)
+            if (sessionExpiryIntervalLimit > 0)
             {
                 mqttPropertyRW.wrap(propertyBuffer, propertiesSize, propertyBuffer.capacity())
-                              .sessionExpiry(sessionExpiryInterval)
+                              .sessionExpiry(sessionExpiryIntervalLimit)
                               .build();
                 propertiesSize = mqttPropertyRW.limit();
             }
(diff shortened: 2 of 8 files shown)


src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1328,6 +1354,16 @@ private void onDecodeConnect(
...
                 case KIND_SESSION_EXPIRY:
+                    if (sessionExpiryInterval != -1)
+                    {
+                        sessionExpiryInterval = 0;
+                        reasonCode = PROTOCOL_ERROR;
+                        break decode;
+                    }
+                    final int sessionExpiryInterval = mqttProperty.expiryInterval();
+                    this.sessionExpiryInterval = Math.min(sessionExpiryInterval, sessionExpiryIntervalLimit);
+                    this.sessionExpiresAt = System.currentTimeMillis() + sessionExpiryInterval * 1000;
```
jfallows: Use `TimeUnit.SECONDS.toMillis(...)` instead of `* 1000` for readability.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -133,6 +140,16 @@
 {
     private static final OctetsFW EMPTY_OCTETS = new OctetsFW().wrap(new UnsafeBuffer(new byte[0]), 0, 0);
 
+    private static final JsonBuilderFactory JSON = Json.createBuilderFactory(new HashMap<>());
+    private static final JsonObjectBuilder OBJECT_BUILDER = JSON.createObjectBuilder();
```
jfallows: These shared instances need to be non-static as they are mutable and not thread-safe.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1398,6 +1435,75 @@ private MqttServerStream resolvePublisher(
...
+
+                    if (mqttConnectPayloadRO.willRetain != 0)
+                    {
+                        willDataFlags = willDataFlags & mqttConnectPayloadRO.willRetain;
+                    }
+
+                    if (mqttConnectPayloadRO.willQos != 0)
+                    {
+                        willDataFlags = willDataFlags & mqttConnectPayloadRO.willQos;
+                    }
```
jfallows: Pass in whatever is needed by the method rather than referring to global shared `mqttConnectPayloadRO` state.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1398,6 +1435,75 @@ private MqttServerStream resolvePublisher(
...
+                                                        .typeId(mqttTypeId)
+                                                        .topic(String.format(SESSION_TOPIC_FORMAT, clientId))
+                                                        .build();
+
+                stream = streams.computeIfAbsent(topicKey, s -> new MqttServerStream(resolvedId, 0, topic));
+
+                final int reserved = payload.sizeof() + mqttConnectPayloadRO.willPayload.sizeof() + stream.initialPadding;
+
+                stream.doApplicationBeginOrFlush(traceId, authorization, affinity, topic, NO_FLAGS, 0, PUBLISH_AND_SUBSCRIBE);
+                stream.doApplicationData(traceId, authorization, reserved, payload, dataEx);
```
jfallows: This needs to be done in two steps, one to extract the `clientId` and initiate the application stream, then another to process the resulting `WINDOW` to trigger sending of the application data, after first acquiring the shared budget for both client and will messages as needed.

If the application stream receives a `RESET`, then we need to handle it the same way as if there is no route, with something like an `AmqpSessionStateStream` that will echo back whatever is published.

src/main/reaktivity/protocol.idl
```diff
@@ -94,6 +95,28 @@
...
+            MqttProperties properties;
+            octets payload;
+        }
+
+        list<uint32, uint32> MqttSession
+        {
+            required uint64 expiresAt;
+            required MqttSubscription[] subscriptions;
+        }
+
```
jfallows: Can now be removed due to JSON formatted session state?

jfallows (changes requested):

src/main/java/org/reaktivity/nukleus/mqtt/internal/MqttConfiguration.java
```diff
@@ -48,6 +49,7 @@
         SUBSCRIPTION_IDENTIFIERS_AVAILABLE = config.property("subscription.identifiers.available", true);
         SHARED_SUBSCRIPTION_AVAILABLE = config.property("shared.subscription.available", false);
         NO_LOCAL = config.property("no.local", false);
+        SESSIONS_AVAILABLE = config.property("sessions.available", true);
```
jfallows: If we configure `session.expiry.interval` to `0`, that would allow us to disable storing session state without needing to add a separate `sessions.available` configuration property, agree?

Though `session.expiry.interval` defaults to `0` now, that will change later, so we would need to explicitly override the configuration to `0` for the ITs to make sure they will continue to work when we support non-zero session expiration interval.

src/main/reaktivity/protocol.idl
```diff
@@ -63,6 +63,7 @@
             string16 value;
         }
 
+
```
jfallows: Please remove unintentional whitespace diff.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1000,6 +1035,7 @@ private boolean hasSubscribeCapability(
         private final long encodeBudgetId;
 
         private final Int2ObjectHashMap<MqttServerStream> streams;
+        private final Int2ObjectHashMap<MqttSessionStateStream> sessionStreams;
```
jfallows: We need just one (wildcard) session stream per connection, agreed?

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1366,6 +1415,14 @@ private void onDecodeConnect(
                 keepAlive = packet.keepAlive();
                 keepAliveTimeout = Math.round(keepAlive * 1.5 * 1000);
                 doSignalKeepAliveTimeoutIfNecessary();
+                if (sessionsAvailable)
```
jfallows: Changing this to be based on `sessionExpiryInterval` has the added benefit of not storing state for sessions that have an expiry of zero, even when we have full support for session state.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1366,6 +1415,14 @@ private void onDecodeConnect(
...
+                {
+                    resolveSession(traceId, authorization, packet.flags(),
```
jfallows: Let's put a block around the `for` loop above and move the label to that block.
```
byte reasonCode = SUCCESS;
decode:
{
    if (connected)
    {
        reasonCode = PROTOCOL_ERROR;
        break decode;
    }
    ...
    for ( ...
}
```
Inside the block, we can `break decode` after setting the `reasonCode` to something other than `SUCCESS`.
After the block, we can have a condition handling non-success.
```
if (reasonCode != SUCCESS)
{
    doNetworkEnd(traceId, authorization);
}
```
This approach will allow local variables such as `mqttConnectPayload` to remain in scope inside the block, avoiding the need to use global variables such as `mqttConnectPayloadRO` instead.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1398,6 +1455,104 @@ private MqttServerStream resolvePublisher(
...
+            final String topic = String.format(SESSION_WILDCARD_TOPIC_FORMAT, clientId);
+            final int topicKey = topicKey(topic);
+
+            final RouteFW route = resolveTarget(routeId, authorization, topic,  PUBLISH_ONLY);
+
+            MqttSessionStateStream stream = sessionStreams.computeIfAbsent(topicKey, s -> new MqttSessionStateStream(0, topic));
+            if (route != null)
+            {
+                final long resolvedId = route.correlationId();
+                stream.initRoute(resolvedId);
```
jfallows: Another possibility is two constructors for `MqttSessionStateStream` instead of `initRoute` where route related fields are also final, with only those needing to be passed being passed to each constructor, depending on `route != null`.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1398,6 +1455,104 @@ private MqttServerStream resolvePublisher(
...
+                                                           .build();
+
+                final OctetsFW sessionPayload = sessionPayloadRW.wrap(payloadBuffer, 0, payloadBuffer.capacity())
+                                                         .set(sessionObj.toString().getBytes(UTF_8))
+                                                         .build();
+
+                stream.doApplicationBeginOrFlush(traceId, authorization, affinity, topic, NO_FLAGS, 0, PUBLISH_AND_SUBSCRIBE);
+
+                final int payloadSize = sessionPayload.sizeof() + willPayload.sizeof();
+
```
jfallows: The logic to test if the session messages can be published will always fail when the stream has just been created, because it won't yet have any stream budget, but it looks like the intent is to re-trigger the decode process when window arrives.

Please confirm that intent, and also that there are no unwanted side effects to running this code multiple times before sufficient budget becomes available.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1328,6 +1367,16 @@ private void onDecodeConnect(
...
                 case KIND_SESSION_EXPIRY:
+                    if (sessionExpiryInterval != -1)
+                    {
+                        sessionExpiryInterval = 0;
+                        reasonCode = PROTOCOL_ERROR;
+                        break decode;
+                    }
+                    final int sessionExpiryInterval = mqttProperty.expiryInterval();
+                    this.sessionExpiryInterval = Math.min(sessionExpiryInterval, sessionExpiryIntervalLimit);
+                    this.sessionExpiresAt = System.currentTimeMillis() + TimeUnit.SECONDS.toMillis(sessionExpiryInterval);
```
jfallows: Given that the session expires a N seconds after the client disconnects abruptly, where session expiry interval is N, is there value to keeping sessionExpiresAt field state in memory?

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1398,6 +1455,104 @@ private MqttServerStream resolvePublisher(
...
+
+            MqttSessionStateStream stream = sessionStreams.computeIfAbsent(topicKey, s -> new MqttSessionStateStream(0, topic));
+            if (route != null)
+            {
+                final long resolvedId = route.correlationId();
+                stream.initRoute(resolvedId);
+
+                final JsonObject sessionObj = objectBuilder.add(SESSION_EXPIRES_AT_NAME, sessionExpiresAt)
+                                                           .add(WILL_DELAY_NAME, willDelay)
+                                                           .build();
```
jfallows: Does it make sense to include `willDelay: 0` if there is no `will` message?

Also, given that the session expires a N seconds after the client disconnects abruptly, this `sessionExpiresAt` value should be updated periodically to make it maximally accurate when the client attempts to reconnect.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -2875,184 +3030,650 @@ private void ensurePublishCapability(
...
+                }
+                else
+                {
+                    if (payload != null)
+                    {
+                        // TODO: get topic
+                        //       decode the payload; check if will or session
+                        //       check session for existing subscriptions, subscribe client to them
+                        // final boolean retainAsPublished = subscription.retainAsPublished();
+                        // doEncodePublish(traceId, authorization, flags, subscription.id, retainAsPublished,
```
jfallows: Given that this wildcard stream is used to manage internal state only, there should be no case where we send `PUBLISH` packets directly to the client representing that internal state.

Inside this `if` block we need a method call, something like `onSessionStateUpdated(payload, extension);` to handle the processing logic.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -2875,184 +3030,650 @@ private void ensurePublishCapability(
...
+                assert reserved >= length + initialPadding;
+
+                initialBudget -= reserved;
+
+                assert initialBudget >= 0;
+
+                sessionStateRO.decodeSessionState(messageType, payload, extension);
+
+                doData(application, routeId, initialId, traceId, authorization, budgetId,
+                    reserved, buffer, offset, length, extension);
```
jfallows: We probably don't need `sessionStateRO` here, and can instead model echo logic based on stream being fully closed or not.
```
if (MqttState.closed(state))
{
    onSessionStateUpdated(payload, extension);
}
else
{
    doData(application, routeId, initialId, traceId, authorization, budgetId,
                    reserved, buffer, offset, length, extension);
}
```

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -2875,184 +3030,650 @@ private void ensurePublishCapability(
...
+            }
+
+            private void cleanup(
```
jfallows: We need to cleanup this session stream from the connection stream when the connection drops or disconnects.

If there is a clean `DISCONNECT` then the session is over and we should publish session state messages to clear out the session state. That would be handled from `onDecodeDisconnect(...)` in the server stream.

jfallows (changes requested):

src/test/java/org/reaktivity/nukleus/mqtt/internal/ConfigurationTest.java
```diff
@@ -22,6 +22,7 @@
 import static org.reaktivity.nukleus.mqtt.internal.MqttConfiguration.NO_LOCAL;
```
jfallows, line 22: Please rename class from `ConfigurationTest` to `MqttConfigurationTest`.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -715,7 +741,7 @@ private int decodePublish(
             else
             {
                 mqttPublishHeaderRO.topic = publish.topicName().asString();
-                mqttPublishHeaderRO.decodeProperties(server, publish.properties());
+                mqttPublishHeaderRO.decode(server, publish.properties());
```
jfallows: Pass in the `publish.topicName()` along with `publish.properties()` instead of modifying a field before the call?
Perhaps return the `reasonCode` directly from `decode(...)`?

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1004,6 +1030,8 @@ private boolean hasSubscribeCapability(
         private final Int2ObjectHashMap<Subscription> subscriptionsByPacketId;
         private final Int2ObjectHashMap<String> topicAliases;
 
+        private MqttSessionStateStream sessionStateStream;
```
jfallows: Perhaps simplify name to `MqttSessionStream` and field to `sessionStream`?

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -660,21 +692,15 @@ else if (!"MQTT".equals(mqttConnect.protocolName().asString()) || mqttConnect.pr
...
                     reasonCode = UNSUPPORTED_PROTOCOL_VERSION;
                 }
-                else if (isSetWillFlag(flags))
-                {
-                    reasonCode = MALFORMED_PACKET;
-                }
                 else if (isSetBasicAuthentication(flags))
                 {
                     reasonCode = NOT_AUTHORIZED;
                 }
```
jfallows, line 707: This section can be hoisted to a separate method and called via `reasonCode = decodeConnectFlags(flags);`

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -660,21 +692,15 @@ else if (!"MQTT".equals(mqttConnect.protocolName().asString()) || mqttConnect.pr
...
             }
 
-            if (reasonCode == 0)
+            if (reasonCode == SUCCESS)
```
jfallows: Let's put the error case first, with `break decode`, then success path follows below `if`.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1293,84 +1322,197 @@ private void doCancelConnectTimeoutIfNecessary()
...
+                        doCancelConnectTimeoutIfNecessary();
+                        doEncodeConnack(traceId, authorization, reasonCode, clientIdentifier.value());
+
+                        connected = true;
+                        keepAlive = connect.keepAlive();
+                        keepAliveTimeout = Math.round(TimeUnit.SECONDS.toMillis(keepAlive) * 1.5);
+                        doSignalKeepAliveTimeoutIfNecessary();
+                        decoder = decodePacketType;
+                        progress = connect.limit();
+                    }
```
jfallows, line 1534: This `else` block duplicates some code above from the `if` block.
Try to bring the codepaths back together and have this code execute less conditionally.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -2257,6 +2441,7 @@ private int calculateSubscribeFlags(
         private final class Subscription
         {
             private int id = 0;
+            private int options = 0;
```
jfallows: Make sure to avoid explicitly initializing non-final member fields to built-in Java defaults, such as `0` for numbers and `null` for objects.

jfallows (changes requested):

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -683,26 +683,24 @@ private int decodeConnect(
                 if ((mqttConnect.typeAndFlags() & 0b1111_1111) != CONNECT_FIXED_HEADER || (flags & CONNECT_RESERVED_MASK) != 0)
                 {
                     reasonCode = MALFORMED_PACKET;
-                }
-                else if (!isSetWillFlag(flags) && (isSetWillQos(flags) || isSetWillRetain(flags)) || invalidWillQos(flags))
-                {
-                    reasonCode = MALFORMED_PACKET;
+                    break decode;
                 }
                 else if (!"MQTT".equals(mqttConnect.protocolName().asString()) || mqttConnect.protocolVersion() != 5)
```
jfallows: Hoist this one to `decodeConnectProtocol` and use a precomputed `StringFW` for comparison instead of using `.asString()`.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -683,26 +683,24 @@ private int decodeConnect(
                 if ((mqttConnect.typeAndFlags() & 0b1111_1111) != CONNECT_FIXED_HEADER || (flags & CONNECT_RESERVED_MASK) != 0)
```
jfallows: Hoist this one to `decodeConnectType`.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -732,17 +730,14 @@ private int decodePublish(
             final MqttPublishFW publish = mqttPublishRO.tryWrap(buffer, offset, offset + server.decodeablePacketBytes);
 
             mqttPublishHeaderRO.reset();
```
jfallows: `final MqttPublishHeader mqttPublishHeader = mqttPublishHeaderRO.reset();`
then
`mqttPublishHeader.decode(...)` instead of `mqttPublishHeaderRO.decode(...)` below.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1030,7 +1025,7 @@ private boolean hasSubscribeCapability(
         private final Int2ObjectHashMap<Subscription> subscriptionsByPacketId;
         private final Int2ObjectHashMap<String> topicAliases;
 
-        private MqttSessionStateStream sessionStateStream;
+        private MqttSessionStateStream sessionStream;
```
jfallows: Suggest renaming class to `MqttSessionStream` also.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1340,55 +1395,11 @@ private int onDecodeConnect(
...
-                    default:
-                        reasonCode = MALFORMED_PACKET;
-                        break decode;
-                    }
-
-                    decodeProgress = mqttProperty.limit();
+                    break decode;
                 }
 
                 final MqttConnectPayload payload = new MqttConnectPayload();
```
jfallows: Suggest sharing and reusing instead of instantiating afresh on each use.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1397,104 +1408,115 @@ private int onDecodeConnect(
 
                 if (sessionExpiryInterval > 0)
                 {
-                    final int flags = connect.flags();
-                    final String topic = String.format(SESSION_WILDCARD_TOPIC_FORMAT, clientId);
-                    final int topicKey = topicKey(topic);
-                    final RouteFW route = resolveTarget(routeId, authorization, topic,  PUBLISH_ONLY);
-
-                    if (route != null)
+                    progress = onResolveSession(traceId, authorization, reasonCode, progress, clientIdentifier, connect, payload);
```
jfallows, line 1436: Suggest renaming to `doEncodeSession(...)`

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1397,104 +1408,115 @@ private int onDecodeConnect(
...
+                }
+                else
+                {
+                    if (reasonCode == SUCCESS)
                     {
-                        final long resolvedId = route.correlationId();
+                        doCancelConnectTimeoutIfNecessary();
+                        doEncodeConnack(traceId, authorization, reasonCode, clientIdentifier.value());
+
+                        connected = true;
```
jfallows: Are we using `connected` to mean two different things?
 - already received `CONNECT`
 - already sent `CONNACK`

Previously, when `onDecodeConnect` called `doEncodeConnack` directly, these were effectively the same, but now that `doEncodeConnack` is asynchronous, we may no longer be properly detecting a second `CONNECT` frame arriving before we have sent `CONNACK`.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1397,104 +1408,115 @@ private int onDecodeConnect(
```
jfallows, line 1428: Should we be doing a `break decode;` here if `reasonCode != SUCCESS` ?

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1397,104 +1408,115 @@ private int onDecodeConnect(
...
-                    final String topic = String.format(SESSION_WILDCARD_TOPIC_FORMAT, clientId);
-                    final int topicKey = topicKey(topic);
-                    final RouteFW route = resolveTarget(routeId, authorization, topic,  PUBLISH_ONLY);
-
-                    if (route != null)
+                    progress = onResolveSession(traceId, authorization, reasonCode, progress, clientIdentifier, connect, payload);
+                }
+                else
+                {
+                    if (reasonCode == SUCCESS)
```
jfallows: If the `break decode;` is added above, then this `if` can be removed.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1397,104 +1408,115 @@ private int onDecodeConnect(
...
+            }
 
-                        final boolean willFlagSet = isSetWillFlag(flags);
+            if (reasonCode != SUCCESS)
+            {
+                doCancelConnectTimeoutIfNecessary();
+                doEncodeConnack(traceId, authorization, reasonCode, clientIdentifier.value());
+                doNetworkEnd(traceId, authorization);
+                progress = connect.limit();
+            }
```
jfallows, line 1458: Does this need to be split into two sections?
  - `if (reasonCode != SUCCESS || sessionExpiryInterval == 0)` => cancel connect timeout + encode connack
  - `if (reasonCode != SUCCESS)` => doNetworkEnd + progress

For the second one, it is sufficient to call `onDecodeError` instead, causing the decoder to switch to ignore-all and for `doNetworkEnd` to be called implicitly?

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1397,104 +1408,115 @@ private int onDecodeConnect(
...
                         }
                     }
-                    else
-                    {
-                        sessionStateStream = new MqttSessionStateStream(0, topic);
-                    }
-                }
-                else
-                {
+
```
jfallows, line 1522: This block can be hoisted to a separate method `decodeWillFlags(connectFlags)`.

jfallows (changes requested):

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1349,28 +1381,162 @@ private void onDecodeConnect(
...
+                    keepAliveTimeout = Math.round(TimeUnit.SECONDS.toMillis(keepAlive) * 1.5);
+                    doSignalKeepAliveTimeoutIfNecessary();
+                    decoder = decodePacketType;
+                    progress = connect.limit();
+                }
             }
 
-            doCancelConnectTimeoutIfNecessary();
-            doEncodeConnack(traceId, authorization, reasonCode, clientIdentifier.value());
+            if (reasonCode != SUCCESS && sessionExpiryInterval <= 0)
```
jfallows: Should this condition be `if (reasonCode != SUCCESS || sessionExpiryInterval == 0)`?
Then we can remove the duplicate code from lines `1427-1428` above.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1398,6 +1564,52 @@ private MqttServerStream resolvePublisher(
             return stream;
         }
 
+        private void resolveSession(
```
jfallows: Suggest we rename this to `doEncodeSession`.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1349,28 +1381,162 @@ private void onDecodeConnect(
...
-                keepAliveTimeout = Math.round(keepAlive * 1.5 * 1000);
-                doSignalKeepAliveTimeoutIfNecessary();
+                doNetworkEnd(traceId, authorization);
+                decoder = decodeIgnoreAll;
+                progress = connect.limit();
+            }
+            return progress;
+        }
+
+        private int doEncodeSession(
```
jfallows: Actually, I think we should revert this name back to `onResolveSession`, and use `doEncodeSession` for `resolveSession` below.

Not sure if it should really return `progress`, but that is minor.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1872,10 +2083,10 @@ private void doEncodeConnack(
         {
             int propertiesSize = 0;
 
-            if (sessionExpiryInterval > 0)
+            if (sessionExpiryIntervalLimit > 0)
             {
                 mqttPropertyRW.wrap(propertyBuffer, propertiesSize, propertyBuffer.capacity())
-                              .sessionExpiry(sessionExpiryInterval)
+                              .sessionExpiry(sessionExpiryIntervalLimit)
```
jfallows: Assign `wrap(...)` return value to local `mqttProperty` variable instead of using `mqttPropertyRW.limit()`.

jfallows (changes requested):

src/test/java/org/reaktivity/nukleus/mqtt/internal/ConnectionIT.java
```diff
@@ -970,8 +1029,24 @@ public void shouldRejectWillRetainWithoutWillFlag() throws Exception
...
+    @Specification({
+        "${routeExt}/session/server/controller",
+        "${client}/connect/session.with.will.message/client",
+        "${server}/publish.session.data/server"})
+    @Configure(name = CLIENT_ID_NAME, value = "one")
+    @Configure(name = WILDCARD_SUBSCRIPTION_AVAILABLE_NAME, value = "true")
+    @Configure(name = SHARED_SUBSCRIPTION_AVAILABLE_NAME, value = "true")
+    @Configure(name = MAXIMUM_QOS_NAME, value = "2")
+    @Configure(name = SESSION_EXPIRY_INTERVAL_NAME, value = "0")
+    public void shouldPublishWillMessageAfterClientDisconnect() throws Exception
```
jfallows: We need two variations of this test, one where the client disconnects abruptly (no `DISCONNECT` packet), which sends the `will message` to the `will topic`, and on where the client disconnects normally (with `DISCONNECT` packet) ending the session, which does not send the `will message`, even though it was specified in the `CONNECT`.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -2875,184 +3102,1418 @@ private void ensurePublishCapability(
...
-        int flags)
-    {
-        return (flags & USERNAME_MASK) != 0;
-    }
+                if (willStream == null)
+                {
+                    this.willStream = new MqttWillStream(sessionStream.routeId, authorization, 0, topic);
+                    willStream.doApplicationBeginOrFlush(traceId, authorization, sessionStream.clientIdentifier, affinity, topic,
+                        NO_FLAGS, 0, PUBLISH_ONLY);
+                }
```
jfallows, line 3120: Can the rest of this method be replaced by a call to `willStream.publishWillMessage(traceId)` instead?

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -2875,184 +3102,1418 @@ private void ensurePublishCapability(
...
+            private void onApplicationReset(
+                ResetFW reset)
+            {
+                setInitialClosed();
+
+                final long traceId = reset.traceId();
+                final long authorization = reset.authorization();
+
+                if (!MqttState.initialOpened(state))
+                {
```
jfallows: needs special behavior for this condition?

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -2875,184 +3102,1418 @@ private void ensurePublishCapability(
...
```
jfallows: Needs special behavior for this condition?

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -2875,184 +3102,1418 @@ private void ensurePublishCapability(
...
+
+                    if (activeStreams.value == 0)
+                    {
+                        activeStreamsByTopic.remove(topicKey);
+                    }
+                }
+            }
+        }
+
+        private class MqttWillStream
```
jfallows, line 3702: Since `will` stream is `publish-only` please remove the methods that are unused, such as on-app-data for the reply.

jfallows (dismissed):

jfallows (dismissed):

jfallows (approved):

Issue #49, opened 2020-05-10
Support shared subscriptions


jfallows:
See MQTT 5 specification:
- section [4.8.2 Shared Subscriptions](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Shared_Subscriptions)

PR #77, opened 2020-10-23
Feature/no local

### LINKED ISSUE #67: Support NO LOCAL flag
See MQTT 5 specification:
- section [3.8.3.1 Subscription Options](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901169)
  - no local

dknoma:
Fix #67 
Requires https://github.com/reaktivity/nukleus-mqtt.spec/pull/49

Added  `SUBSCRIBE_ONLY` -> `PUBLISH_AND_SUBSCRIBE` no-local and `PUBLISH_ONLY` -> `PUBLISH_AND_SUBSCRIBE` no-local scenarios. PUBLISH_ONLY` -> `PUBLISH_AND_SUBSCRIBE` also now properly encodes and sends `SUBACK` to the client.

### DIFF
--- a/src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
+++ b/src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
@@ -2488,7 +2488,6 @@ public final class MqttServerFactory implements StreamFactory
             private int ackCount;
             private int successMask;
             private int ackMask;
-            private boolean acknowledged;
 
             private Subscription()
             {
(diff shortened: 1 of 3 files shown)


src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -2558,20 +2559,19 @@ private boolean retainAsPublished()
 
             private int state;
             private int capabilities;
+            private boolean didSendSuback;
```
jfallows: Lets move this to `subscription.acknowledged` instead.

jfallows (changes requested):

jfallows (dismissed):

jfallows (approved):

PR #84, opened 2020-12-08
Support Session Resumption

dknoma:
Implement full session support in `MQTT`. Includes:

- Session expiration
- Support clean start and no clean start
- Will message delivery
- Subscription resumption upon session recovery


src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -2607,18 +2607,22 @@ private void onSubscribeSucceeded(
...
             }
 
-            private void onSubscribeCompleted(
+            private boolean onSubscribeCompleted(
                 long traceId,
                 long authorization,
                 int packetId)
             {
+                boolean completed = false;
                 if (Integer.bitCount(ackMask) == ackCount)
```
jfallows: Let's move this to a separate method on `Subscription` such as:
```
private boolean acknowledged()
{
    return Integer.bitCount(ackMask) == ackCount;
}
```

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -2906,7 +2910,11 @@ private void onApplicationWindow(
                 if (subscription != null &&
                         !subscription.hasSubscribeCompleted(subackIndex) && hasSubscribeCapability(capabilities))
                 {
-                    subscription.onSubscribeSucceeded(traceId, authorization, packetId, subackIndex);
+                    final boolean completed = subscription.onSubscribeSucceeded(traceId, authorization, packetId, subackIndex);
+                    if (completed)
```
jfallows: Avoid chaining the `boolean` `completed` flag through the methods and instead use the new `subscription.acknowledged()` method for the `if` condition.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -3025,7 +3033,11 @@ private void onApplicationBegin(
                 final long traceId = begin.traceId();
                 final long authorization = begin.authorization();
 
-                doApplicationWindowIfNecessary(traceId, authorization);
+                if (hasPublishCapability(capabilities))
+                {
+                    doApplicationWindowIfNecessary(traceId, authorization);
+                }
+                // doApplicationWindowIfNecessary(traceId, authorization);
```
jfallows: Remove commented?

src/test/java/org/reaktivity/nukleus/mqtt/internal/ConnectionIT.java
```diff
@@ -1130,4 +1130,18 @@ public void shouldConnectWithWillMessageThenPublishOneMessage() throws Exception
...
+    //     "${client}/subscribe.one.message.suback.order/client",
+    //     "${server}/subscribe.one.message.suback.order/server"})
+    // @Configure(name = WILDCARD_SUBSCRIPTION_AVAILABLE_NAME, value = "true")
+    // @Configure(name = SHARED_SUBSCRIPTION_AVAILABLE_NAME, value = "true")
+    // @Configure(name = MAXIMUM_QOS_NAME, value = "2")
+    // @Configure(name = SESSION_EXPIRY_INTERVAL_NAME, value = "0")
+    // public void shouldSubscribeOneMessageSubackOrder() throws Exception
+    // {
+    //     k3po.finish();
+    // }
```
jfallows: Should this be uncommented or removed?

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -2906,7 +2910,11 @@ private void onApplicationWindow(
                 if (subscription != null &&
                         !subscription.hasSubscribeCompleted(subackIndex) && hasSubscribeCapability(capabilities))
                 {
-                    subscription.onSubscribeSucceeded(traceId, authorization, packetId, subackIndex);
+                    final boolean completed = subscription.onSubscribeSucceeded(traceId, authorization, packetId, subackIndex);
+                    if (completed)
+                    {
+                        doApplicationWindow(traceId, authorization);
+                    }
```
jfallows, line 2958: Shouldn't this conditional sending of `window` be done in `onApplicationBegin` instead?

Also, you need to handle the case where `begin` arrived for multiple reply streams, and the first one couldn't send `window` because the `subscription` had not yet been `acknowledged`, then the last reply stream comes in, allowing the `subscription` to be `acknowledged` so all application streams for the `subscription` should now be notified to send `window` if necessary otherwise they would be stalled.

jfallows (changes requested):

PR #87, opened 2020-12-19
Fix sending of Will Message over application

dknoma:
Fix possible null pointer exceptions during will message encoding.

Fix sessionExpiryInterval property inappropriately being encoded in `CONNACK` under certain circumstances.

spec: https://github.com/reaktivity/nukleus-mqtt.spec/pull/56

### DIFF
--- a/src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
+++ b/src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
@@ -183,6 +183,8 @@ public final class MqttServerFactory implements StreamFactory
     private static final int CONNECT_TOPIC_ALIAS_MAXIMUM_MASK = 0b0000_0001;
     private static final int CONNECT_SESSION_EXPIRY_INTERVAL_MASK = 0b0000_0010;
 
+    private static final int CONNACK_SESSION_PRESENT = 0b0000_0001;
+
     private static final int RETAIN_HANDLING_SEND = 0;
 
     private static final int RETAIN_FLAG = 1 << RETAIN.ordinal();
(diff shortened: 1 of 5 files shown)


src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -2263,12 +2275,14 @@ private void doEncodeConnack(
...
 
+            int flags = 0x01;
+
             final int propertiesSize0 = propertiesSize;
             final MqttConnackFW connack =
                     mqttConnackRW.wrap(writeBuffer, FIELD_OFFSET_PAYLOAD, writeBuffer.capacity())
                                        .typeAndFlags(0x20)
                                        .remainingLength(3 + propertiesSize0)
-                                       .flags(0x00)
+                                       .flags(flags)
```
jfallows, line 2287: Let's use a constant like `CONNACK_SESSION_PRESENT` instead of `flags` for improved readability.

src/main/java/org/reaktivity/nukleus/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -2199,7 +2211,7 @@ private void doEncodeConnack(
             int propertiesSize = 0;
 
             MqttPropertyFW mqttProperty;
-            if (sessionExpiryIntervalLimit > 0)
+            if (sessionExpiryInterval > sessionExpiryIntervalLimit && sessionExpiryIntervalLimit > 0)
```
jfallows: What if we made the default limit `unlimited`, i.e. `Integer.MAX_VALUE`, then this logic could revert to the simpler form of `if (sessionExpiryInterval > sessionExpiryIntervalLimit)`, agree?

src/test/java/org/reaktivity/nukleus/mqtt/internal/ConnectionIT.java
```diff
@@ -18,6 +18,7 @@
 import static java.util.concurrent.TimeUnit.SECONDS;
 import static org.junit.rules.RuleChain.outerRule;
 import static org.reaktivity.nukleus.mqtt.internal.MqttConfiguration.PUBLISH_TIMEOUT;
+import static org.reaktivity.nukleus.mqtt.internal.MqttConfigurationTest.CLIENT_ID_NAME;
```
jfallows: Unused import?

jfallows (changes requested):

jfallows (approved):

